PR #4, opened 2020-03-11
Restructure Use Cases

wiresio:


mmccool:
Merging, useful reorg, but we will probably want to break detailed use cases out into separate files to make them easier to copy over to wot-architecture.

PR #20, opened 2020-04-27
UPM Use Case about Smart Campus

AndreaCimminoArriaga:
Universidad Politécnica de Madrid (UPM) use case for wot-discovery about cross-domain discovery in a smart campus


mmccool:
Please do the following:
1. Please fix IPR issue by merging W3C/github accounts with Kaz's guidance
2. We will merge the above... but we need to capture the "remote directory" requirement and generate a PR to merge this UC with the other ones in wot-architecture

PR #35, opened 2020-07-06
Initial WoT Discovery index.html outline

mmccool:
An initial index.html for the WoT Discovery deliverable, with a proposed outline.

***
<a href="https://pr-preview.s3.amazonaws.com/mmccool/wot-discovery/pull/35.html" title="Last updated on Jul 20, 2020, 2:53 PM UTC (2b47d6f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-discovery/35/e22550e...mmccool:2b47d6f.html" title="Last updated on Jul 20, 2020, 2:53 PM UTC (2b47d6f)">Diff</a>


index.html
```diff
@@ -0,0 +1,200 @@
...
+</head>
+<body>
+    <section id="abstract">
+        <p>The W3C Web of Things (WoT) is intended to enable
+            interoperability across IoT platforms and application
+            domains.
+            Overall, the goal of the WoT is to preserve and complement existing
+            IoT standards and solutions. In general, the W3C WoT architecture is 
+            designed to describe what exists rather than to prescribe
+            what to implement.
```
benfrancis, line 69: FWIW I've never really agreed with this statement from the architecture document, but it seems particularly contradictory to include it in the abstract of a specification which it appears will then go on to describe a Directory Service API which WoT clients and servers should implement.

mmccool:
Comments from meeting:
- Different formats for search results (eg RDF, JSON, JSON-LD)
- Different query types (eg SPARQL vs JSONPath vs XPath)
- For JSONPath/XPath queries, when is return data JSON-LD and when is it just JSON?
- Should we include an OpenAPI specification for the API?   There are things like error codes and header options that can't currently be specified in TDs.  BUT if we have two different specs for the directory there is a chance of inconsistency.   Might be better to define extensions to add the necessary information to a TD.
- Additional exploration types, in particular direct-to-device; constraint is that some introductions only give an address and no type, so we have to be able to tell them apart AFTER going to the exploration service
- Need to decide what introduction types should be included in the spec: 
     - Direct URL (also includes printed QR codes, bluetooth beacons, etc.)
     - DNS-SD (needs a service type)
     - CoRE RD (needs a link relation type)
     - DID Document (needs a link relation type; same as one in CoRE RD)
- How do we deal with RDF summaries?  Perhaps associated with the link pointing at another directory [Andrea will follow up]
- Minimum security requirements for directory API (special section)
- Listing vs. Search, e.g. Search is list + filtering

mmccool:
Resolution (in Directory TF call): add Farshid Tavakolizadeh, Kunihiko Toumura, and Andrea Cimmino as co-editors

index.html
mmccool: Yeah, that was just cut-and-paste.   I'm going to rewrite this section.

mmccool:
Not quite sure why preview fails to work, it works fine if I preview it locally.

index.html
```diff
@@ -0,0 +1,321 @@
...
+            <h2>Direct</h2>
+            <p>Any mechanism that results in a single URL.
+               This includes Bluetooth beacons, QR codes, and written URLs to be 
+               typed by a user.
+               A GET on all such URLs MUST result in a TD.
+               For self-describing Things, this can be the TD of the Thing itself.
+               If the URL references a Directory, this MUST be the TD of the
+               Directory service.
+               A Directory can be distinguished from a Thing by the use of an
+               <code>@type</code> including the semantic term <code>WoT-Directory</code>.
```
farshidtz, line 214: I believe lines 209-214 describe the beginning of the exploration phase and are subject to access control. If that is correct, it will not be possible to distinguish a Directory from a Thing within Direct Introduction without something distinctive inside the URL. I remember discussing well-known paths for this purpose.

ashimura:
Regarding the PR-Preview error 500, I suspect PR-Preview requires an initial version as the basis of its diff generation capability. So I've pushed a [template index.html](https://github.com/w3c/wot-discovery/blob/master/index.html) to the wot-discovery repo and suggest @mmccool regenerate this PR or simply push the initial edit as the basis of the discussion.

k-toumura:
Here is a [rendered version](https://cdn.statically.io/gh/w3c/wot-discovery/3f13162d382d5ba9257a438dc69abecec4e96808/index.html?env=dev) for review.

index.html
mmccool: That is a useful point, but... I can imagine introductions that need authentication, so

mmccool: we can resolve in future PRs.   Needs more discussion

PR #36, opened 2020-07-09
Thing Description for LinkSmart directory

farshidtz:
This is an improved version of the TD [presented in F2F](https://github.com/w3c/wot/blob/master/PRESENTATIONS/2020-06-online-f2f/2020-06-22-Discovery-ThingDirectory-Tavakolizadeh.pdf) for the LinkSmart directory.

### DIFF
--- a/prior-work/fraunhofer/directory-td.jsonld
+++ b/prior-work/fraunhofer/directory-td.jsonld
@@ -11,8 +11,18 @@
     },
     "security": "bearer_sc",
     "actions": {
-      "td": {
-        "description": "Add, modify, or remove a Thing Description",
+      "postTD": {
+        "description": "Add an anonymous Thing Description (without ID)",
+        "forms": [
+          {
+            "href": "/td",
+            "htv:methodName": "POST",
+            "contentType": "application/ld+json"
+          }
+        ]
+      },
+      "putTD": {
+        "description": "Add or modify a Thing Description",
         "uriVariables": {
           "id": {
             "title": "Thing Description ID",
@@ -22,20 +32,39 @@
         },
         "forms": [
           {
-            "href": "/td",
-            "htv:methodName": "POST",
-            "contentType": "application/ld+json"
-          },
-          {
             "href": "/td/{id}",
             "htv:methodName": "PUT",
             "contentType": "application/ld+json"
-          },
+          }
+        ]
+      },
+      "patchTD": {
+        "description": "Modify parts of a Thing Description",
+        "uriVariables": {
+          "id": {
+            "title": "Thing Description ID",
+            "type": "string",
+            "format": "iri-reference"
+          }
+        },
+        "forms": [
           {
             "href": "/td/{id}",
             "htv:methodName": "PATCH",
             "contentType": "application/ld+json"
-          },
+          }
+        ]
+      },
+      "deleteTD": {
+        "description": "Remove a Thing Description",
+        "uriVariables": {
+          "id": {
+            "title": "Thing Description ID",
+            "type": "string",
+            "format": "iri-reference"
+          }
+        },
+        "forms": [
           {
             "href": "/td/{id}",
             "htv:methodName": "DELETE"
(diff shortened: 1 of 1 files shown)


prior-work/fraunhofer/directory-td.jsonld
```diff
@@ -0,0 +1,98 @@
...
+            "href": "/td/{id}",
+            "htv:methodName": "PATCH",
+            "contentType": "application/ld+json"
+          },
+          {
+            "href": "/td/{id}",
+            "htv:methodName": "DELETE"
+          }
+        ]
+      }
```
benfrancis, line 44: This is a bit of a strange example of an action because it's really several different actions, as I think is revealed by the ambiguous name. I would usually expect an action to be describable with a verb or verb phrase.

prior-work/fraunhofer/directory-td.jsonld
```diff
@@ -0,0 +1,98 @@
...
+          }
+        },
+        "forms": [
+          {
+            "href": "/td/{id}",
+            "htv:methodName": "GET",
+            "contentType": "application/ld+json"
+          }
+        ]
+      },
```
benfrancis, line 63: Using URL templates in the form of a property description is also kind of odd, because it's really a collection of multiple resources which is being described rather than a single property resource.

prior-work/fraunhofer/directory-td.jsonld
farshidtz: I had my concerns when mapping the [OpenAPI specs](https://github.com/linksmart/thing-directory/blob/master/apidoc/openapi-spec.yml) to TD ([see the last slide](https://github.com/w3c/wot/blob/master/PRESENTATIONS/2020-06-online-f2f/2020-06-22-Discovery-ThingDirectory-Tavakolizadeh.pdf)). 

I agree that they are different actions, and separating them makes it clearer. However, I would keep the names equivalent to the HTTP methods, because they have their own semantics.
```json
{
  ...
  "actions": {
    "postTD": {
      "description": "Add an anonymous Thing Description (without ID)",
      "forms": [
        {
          "href": "/td",
          "htv:methodName": "POST",
          "contentType": "application/ld+json"
          // how to specify that the response contains a system-generated id in Location header?
        }
      ]
    },
    "putTD": {
      "description": "Add or modify a Thing Description",
      "uriVariables": {
        "id": {
          "title": "Thing Description ID",
          "type": "string",
          "format": "iri-reference"
        }
      },
      "forms": [
        {
          "href": "/td/{id}",
          "htv:methodName": "PUT",
          "contentType": "application/ld+json"
        }
      ]
    },
    "patchTD": {
      "description": "Modify parts of a Thing Description",
      "uriVariables": {
        "id": {
          "title": "Thing Description ID",
          "type": "string",
          "format": "iri-reference"
        }
      },
      "forms": [
        {
          "href": "/td/{id}",
          "htv:methodName": "PATCH",
          "contentType": "application/ld+json"
        }
      ]
    },
    "deleteTD": {
      "description": "Remove a Thing Description",
      "uriVariables": {
        "id": {
          "title": "Thing Description ID",
          "type": "string",
          "format": "iri-reference"
        }
      },
      "forms": [
        {
          "href": "/td/{id}",
          "htv:methodName": "DELETE"
        }
      ]
    }
  }
  ...
}
```

prior-work/fraunhofer/directory-td.jsonld
farshidtz: IMO, TD has evolved for describing things and not collection services. Trying to use a TD to describe a collection API has its side effects. I decided to map "properties" to resources of the collection. I also thought about "getTD" and "listTD" as actions on the collection, but it seems contradictory to the specs because those operations don't manipulate the state of the collection.

"An Interaction Affordance that allows to invoke a function of the Thing, which **manipulates state**" [[1](https://www.w3.org/TR/wot-thing-description/#actionaffordance)]

benfrancis: I agree that getTD and ListTD don't make sense as actions.

PR #38, opened 2020-07-25
Draft of registration API

farshidtz:
Added a draft for the Registration API.

- Not sure if it should be more specific and also specify the API endpoint specifications or just point to the action/property name of the directory TD. 
- Should TD's `id` be mutable inside a directory? Apart from privacy risks (see #39), this complicates the API design.

***
<a href="https://pr-preview.s3.amazonaws.com/farshidtz/wot-discovery/pull/38.html" title="Last updated on Aug 4, 2020, 9:54 AM UTC (97c2123)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-discovery/38/5ff4fe9...farshidtz:97c2123.html" title="Last updated on Aug 4, 2020, 9:54 AM UTC (97c2123)">Diff</a>

### DIFF
--- a/directory.td.json
+++ b/directory.td.json
@@ -1,0 +1,186 @@
+{
+    "@context": "https://www.w3.org/2019/wot/td/v1",
+    "@type": "WoT-Directory", // namespace?
+    "title": "Thing Description Directory (TDD)",
+    "securityDefinitions": {
+        "oauth2_code": {
+            "scheme": "oauth2",
+            "flow": "code",
+            "authorization": "https://auth.example.com/authorization",
+            "token": "https://auth.example.com/token",
+            "scopes": [
+                "write",
+                "read",
+                "search"
+            ]
+        }
+    },
+    "security": "oauth2_code",
+    "base": "https://tdd.example.com",
+    "actions": {
+        "createTD": {
+            "description": "Add a Thing Description",
+            "uriVariables": {
+                "id": {
+                    "title": "Thing Description ID",
+                    "type": "string",
+                    "format": "iri-reference"
+                }
+            },
+            "forms": [
+                {
+                    "href": "/td/{id}", // use /registrations instead?
+                    "htv:methodName": "PUT",
+                    "contentType": "application/td+json",
+                    "response": {
+                        "description": "Success response", // need to define vocab?
+                        "htv:statusCodeValue": 201
+                    },
+                    "scopes": "write"
+                },
+                {
+                    "href": "/td",
+                    "htv:methodName": "POST",
+                    "contentType": "application/td+json",
+                    "response": {
+                        "description": "Success response", // need to define vocab?
+                        "htv:headers": [
+                            {
+                                "description": "System-generated UUID (version 4) URN", // need to define vocab?
+                                "htv:fieldName": "Location",
+                                "htv:fieldValue": ""
+                            }
+                        ],
+                        "htv:statusCodeValue": 201
+                    },
+                    "scopes": "write"
+                }
+            ]
+        },
+        "updateTD": {
+            "description": "Modify an Thing Description",
+            "uriVariables": {
+                "id": {
+                    "title": "Thing Description ID",
+                    "type": "string",
+                    "format": "iri-reference"
+                }
+            },
+            "forms": [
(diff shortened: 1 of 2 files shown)


farshidtz:
Added the draft of TD for directory as a JSON file. It is easier to work on it this way, will add to index.html once stable.

-  There is a `Thing` type in https://www.w3.org/2019/wot/td/v1, is a `Directory` (or WoT-Directory as in index.html) type going to be added there?
- No easy way to describe multiple response types, unless response array gets added to the spec (https://github.com/w3c/wot-thing-description/issues/617)
- Used `description` inside the `response`; is this allowed?
- Used `/td` as root endpoint of the registration API. `/registrations` is another option; plural because GET /registrations is expected to respond with all TDs.
- Used OAuth 2.0 code flow for now, since this is the only one currently defined in the TD spec. This is just an example and we need to include client, device, code with an OR scheme.

farshidtz:
Changes based on the comments during the discovery call:
- changed `@type` from `WoT-Directory` to `Directory`
- `htv` is already included in https://www.w3.org/2019/wot/td/v1, so we don't need to add it again in the current context.
- added editor's notes for ID type, scope, [normative code block](#41)

benfrancis:
Some quick feedback:
1. I don't think this API design for adding web things to a directory would fulfill either of the two key use cases Mozilla's WebThings Gateway has for a WoT directory, which are 1) Add an existing web thing by its URL and 2) Add a non-WoT device which is then hosted as a web thing by the directory with an automatically generated Thing Description. I have filed #45 and #46 to explain these use cases in more depth.
2. I find it a bit strange that create, update and delete operations are actions and the read operation is a property but the forms of each interaction affordance all have the same URI in their href member. If you read a property to get a Thing Description why would you not write a property to update it? I don't think this is the fault of the author but rather a sign that Thing Descriptions are not well suited to describing collections of resources and I'm no longer convinced exposing a TD for a directory is a good idea.

farshidtz:
Thank you for the feedback.

> 1. I don't think this API design for adding web things to a directory would fulfill either of the two key use cases Mozilla's WebThings Gateway has for a WoT directory, which are 1) Add an existing web thing by its URL and 2) Add a non-WoT device which is then hosted as a web thing by the directory with an automatically generated Thing Description. I have filed #45 and #46 to explain these use cases in more depth.

Such functionalities can be added to the directory using other microservices or extensions. They don't need to be a part of a TD registration API.
Fetching a resource and maintaining it internally, as mentioned before, is subject to network restrictions and access control. Automatically generating TDs is application specific and requires adapters as you explained [here](https://github.com/w3c/wot-discovery/issues/46). 

> 2. I find it a bit strange that create, update and delete operations are actions and the read operation is a property but the forms of each interaction affordance all have the same URI in their href member. If you read a property to get a Thing Description why would you not write a property to update it? I don't think this is the fault of the author but rather a sign that Thing Descriptions are not well suited to describing collections of resources and I'm no longer convinced exposing a TD for a directory is a good idea.

I think there is conflicting information in the [TD spec](https://www.w3.org/TR/wot-thing-description/):
> Action: An Interaction Affordance that allows to invoke a function of the Thing, **which manipulates state** (e.g., toggling a lamp on or off) or triggers a process on the Thing (e.g., dim a lamp over time).

> Property: An Interaction Affordance that exposes state of the Thing. This state can then be retrieved (read) and **optionally updated** (write).

Action is for manipulating state of the Thing, in this case the Thing is the directory and its resources are being manipulated. I don't know why at the same time setting a property is allowed by the TD spec.

mmccool:
Regarding 1, I agree with Farshid: we are giving directory services a narrow scope on purpose (as a microservice) under the assumption that there would be other microservices that would handle the two use cases you mention.  In particular it would be useful to have a "discoverer" service whose job it is to "find" Things and register them, and conversely protocol translation would be a separate service, which in turn would register generated TDs with the directory.   So IMO the current spec is not in conflict with these use cases, although it is true it does not "cover" them (and maybe we should at least discuss this in the spec, eg in the architecture section).

mmccool:
I think we should merge this PR and iterate with further issues and PRs.

PR #47, opened 2020-08-10
Updated index.html with search

AndreaCimminoArriaga:
***
<a href="https://pr-preview.s3.amazonaws.com/AndreaCimminoArriaga/wot-discovery/pull/47.html" title="Last updated on Sep 23, 2020, 8:48 AM UTC (d718c18)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-discovery/47/bd8e7c9...AndreaCimminoArriaga:d718c18.html" title="Last updated on Sep 23, 2020, 8:48 AM UTC (d718c18)">Diff</a>


mmccool:
- SPARQL support should be optional
- sub-API should be defined in a way that can be easily separated (maybe a separate Thing?)
- Return is not a TD, since might be a fragment, etc.  How to return a full TD?  When will the content type be the TD content type?   Discussion: partial returns for SELECT; same issue for XPath, etc; but if don't use SELECT, but DESCRIBE, can still extract TDs... output will be a TD, as triples, but can ask result to be in JSON-LD, so...
- We need two implementations... are there two SPARQL databases we can use that are up to date?

mmccool:
Also, for numbering sections: use header tags with <section> for ReSpec. See also: https://github.com/w3c/respec/wiki/section

AndreaCimminoArriaga:
* SPARQL support should be optional
    * After a discussion with @farshidtz , I agree on this
* sub-API should be defined in a way that can be easily separated (maybe a separate Thing?)
    * I though on having on the /td API the parameter /td?query=XXX to support the SPARQL (this is compliant with the current standard). Maybe we can have /td/sparql or directly /sparql as separate API?
* Return is not a TD, since might be a fragment, etc. How to return a full TD? When will the content type be the TD content type? Discussion: partial returns for SELECT; same issue for XPath, etc; but if don't use SELECT, but DESCRIBE, can still extract TDs... output will be a TD, as triples, but can ask result to be in JSON-LD, so...
    * SELECT queries can definitively return fragments of the TD based on the variables defined in the query. Regarding the whole TD, this can be achieved using [DESCRIBE](https://www.w3.org/TR/rdf-sparql-query/#describe) query.

* We need two implementations... are there two SPARQL databases we can use that are up to date?
    * There are several triple stores that can be used (GraphDB, Virtuoso, Jena Fuseky, ...)

mmccool:
Discussion in discovery call:
- Let's put different query styles in different URL sub-trees, i.e. /td/sparql, /td/xpath, etc. to avoid problems
- How to deal with different responses with different content types?   Do we need multiple "responses"/"alternateResponses" for this?

farshidtz:
After some thought, I think `GET /td/sparql` is not really RESTful as it indicates retrieval of a sparql resource. 

I still prefer the following:
- `GET /td?sparql=xxx` -> Retrieving TDs after applying sparql xxx to TDs
- `GET /td?jsonpath=yyy` -> Retrieving TDs after after applying jsonpath yyy to TDs

This allows extension to:
- `GET /td` -> Retrieving all TDs, in paginated form (default page=1). 
- `GET /td?page=2&per_page=100` -> Retrieving 2nd page of all TDs.  Or use offset/limit. Maximum value for limit/per_page should be decided by server. Server can also recommend those in nextLink.

Only one of sparql, jsonpath, (page, page_page) shall be allowed.

The POST variant for sparql could be:
- `POST /td/sparql` -> Submitting sparql and getting responses

If not ideal, we could look into an alternate design with `/search` as root endpoint.

index.html
```diff
@@ -678,8 +678,59 @@ <h4>Search</h4>
...
+		    <span class="rfc2119-assertion" id="tdd-search-apis">
+			The Directory MUST implement the syntactic search with JSON Path, SHOULD implement the syntactic search with XPath, and MAY implement semantic search with SPARQL.
+		    </span>
+        <p class="ednote" title="Normative JSON Block">Should we have the previous sentence splitted into three different assertions?.</p>
+    		
+        <section id="jsonpath-semantic" class="normative">
+          <h4>Syntactic search: JSONPath</h4>
+          <span class="rfc2119-assertion" id="tdd-search-jsonpath-method"> The API MUST allow searching TDs using an HTTP <code>GET</code> request.</span> 
+          <span class="rfc2119-assertion" id="tdd-search-jsonpath-parameter"> The request MUST contain a valid JSON Path [[JSON Path]] as searching parameter.</span>
+          <span class="rfc2119-assertion" id="tdd-search-jsonpath-response">  A successful response MUST have 200 (OK) status, contain <code>application/json</code> Content-Type header, and in the body a set of complete TDs or a set of TD elements.</span>
```
farshidtz, line 692: As suggested in discovery call, we can define another global term to define "TD Fragments" (in contrast to Partial TDs). This PR refers to this as TD elements in few places. This can be added in a separate PR.

index.html
```diff
@@ -678,8 +678,59 @@ <h4>Search</h4>
...
+    	  <ul>
+    	   <li>400(Bad Request): if the provided <code>XPath</code> contains syntax errors.</li>
+           <li>401 (Unauthorized): No authentication.</li>
+           <li>403 (Forbidden): Insufficient rights to the resource.</li>
+    	  </ul>
+    	 </p>
+        </section>
+	<section id="search-semantic" class="normative">
+      	 <h4>Semantic search: SPARQL</h4>
+      	 <span class="rfc2119-assertion" id="tdd-search-semantic-method"> The API MUST allow searching TDs using the SPARQL protocol [[sparql11-overview]]. The API MUST implement the protocol using <code>GET</code> requests, and MAY implement the protocol for <code>POST</code> requests.</span>
```
farshidtz, line 720: The three assertions should probably be separated into three assertion blocks.

farshidtz (changes requested):
I added several inline comments.

Some general comments:
- There are several uses of "JSON Path" and "Json Path" (in TD and index.html) in place of "JSONPath"
- 400 (Bad Request) may also be used when `query` parameter is not given.
- JSONPath/XPath/SPARQL are still in `code` tag in error response sections.

mmccool:
Deferring merge until rebasing (and my review).

directory.td.json
```diff
@@ -1,153 +1,225 @@
...
+      "uriVariables": {
+        "query": {
+          "title": "A valid Json Path expression",
+          "type": "string",
+          "format": "iri-reference"
+        }
+      },
+      "forms": [
+        {
+          "href": "/search/jsonpath?query={expression}",
```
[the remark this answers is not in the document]
AndreaCimminoArriaga: Fixed in last commit (d925febf31f0315ee9c9a06909e518779f9dbfc4)

directory.td.json
```diff
@@ -1,153 +1,225 @@
...
+      "uriVariables": {
+        "query": {
+          "title": "A valid XPath expression",
+          "type": "string",
+          "format": "iri-reference"
+        }
+      },
+      "forms": [
+        {
+          "href": "/search/xpath?query={expression}",
```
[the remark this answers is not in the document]
AndreaCimminoArriaga: Fixed in commit d925febf31f0315ee9c9a06909e518779f9dbfc4

index.html
```diff
@@ -678,8 +678,59 @@ <h4>Search</h4>
...
                     <li>Spatial and network-limited queries</li>
                     <li>Links</li>
-                    </ul>
-                </section>
+		    </ul>
+        The Directory has three search APIs: syntactic search with JSON Path [[JSON Path]] or XPath [[xpath-31]], and semantic search with SPARQL [[sparql11-overview]].
+		    <span class="rfc2119-assertion" id="tdd-search-apis">
+			The Directory MUST implement the syntactic search with JSON Path, SHOULD implement the syntactic search with XPath, and MAY implement semantic search with SPARQL.
+		    </span>
+        <p class="ednote" title="Normative JSON Block">Should we have the previous sentence splitted into three different assertions?.</p>
```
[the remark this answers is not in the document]
AndreaCimminoArriaga: Solved in commit d718c1868e4a604a176ea72629ec4f686f0341ab

index.html
AndreaCimminoArriaga: I changed the term in commit d718c1868e4a604a176ea72629ec4f686f0341ab

index.html
AndreaCimminoArriaga: I separated the assertion into three blocks as suggested in commit d718c1868e4a604a176ea72629ec4f686f0341ab

PR #50, opened 2020-08-15
Add additional security schemes to directory TD

farshidtz:
Added client and device security schemes based on the proposed combination scheme (https://github.com/w3c/wot-thing-description/pull/944). 

While adding those, I started to have doubts in the recently merged spec (https://github.com/w3c/wot-thing-description/pull/927) about what endpoints are necessary for code and device flows.

In this PR, I included what is technically **required** by a user to interact with a thing based on the TD. In particular: 
- In code flow, the user only needs the `authorization` endpoint. The `token` endpoint is required by the device to internally exchange the code with an access token. Why MUST the `token` endpoint be included and given to the consumer in device's metadata?
- In device flow, the device talks to `authorization` (oauth2 device authorization) endpoint internally, gets a `verification_uri` and `user_code` from the server and displays these to the user. The user follows the `verification_uri` to authenticate/authorize. At the same time, the device starts polling the `token` endpoint to receive the access token. Why MUST the `authorization` and `token` endpoints be included and given to the consumer in device's metadata?

***
<a href="https://pr-preview.s3.amazonaws.com/farshidtz/wot-discovery/pull/50.html" title="Last updated on Aug 18, 2020, 5:37 PM UTC (e8470f2)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-discovery/50/d964f85...farshidtz:e8470f2.html" title="Last updated on Aug 18, 2020, 5:37 PM UTC (e8470f2)">Diff</a>

### DIFF
--- a/directory.td.json
+++ b/directory.td.json
@@ -10,15 +10,41 @@
             "scheme": "oauth2",
             "flow": "code",
             "authorization": "https://auth.example.com/authorization",
+            "scopes": [
+                "write",
+                "read",
+                "search"
+            ]
+        },
+        "oauth2_client": {
+            "scheme": "oauth2",
+            "flow": "client",
             "token": "https://auth.example.com/token",
             "scopes": [
                 "write",
                 "read",
                 "search"
             ]
+        },
+        "oauth2_device": {
+            "scheme": "oauth2",
+            "flow": "device",
+            "scopes": [
+                "write",
+                "read",
+                "search"
+            ]
+        },
+        "combination": {
+            "scheme": "combination",
+            "oneOf": [
+                "oauth2_code",
+                "oauth2_client",
+                "oauth2_device"
+            ]
         }
     },
-    "security": "oauth2_code",
+    "security": "combination",
     "base": "https://tdd.example.com",
     "actions": {
         "createTD": {
--- a/directory.td.json
+++ b/directory.td.json
@@ -10,6 +10,7 @@
             "scheme": "oauth2",
             "flow": "code",
             "authorization": "https://auth.example.com/authorization",
+            "token": "https://auth.example.com/token",
             "scopes": [
                 "write",
                 "read",
@@ -29,6 +30,8 @@
         "oauth2_device": {
             "scheme": "oauth2",
             "flow": "device",
+            "authorization": "https://auth.example.com/device_authorization",
+            "token": "https://auth.example.com/token",
             "scopes": [
                 "write",
                 "read",


relu91:
> In code flow, the user only needs the authorization endpoint. The token endpoint is required by the device to internally exchange the code with an access token. Why MUST the token endpoint be included and given to the consumer in device's metadata?

We may expand this on today's call. However, based on my assumptions on the use case we have the following actors:
- The resource owner (let's say it is a human)
- The resource (in this case one of the affordances exposed by the TTD)
- The device which wants to access such resource. 

Following the diagram below:
```
                                                 +-----------+
  +----------+                                   |           |
  | Resource |                                   |  TDD      |
  |   Owner  |                                   |           |+<-------+
  |          |                                   |           |        |
  +----+-----+                                   +-----------+        |
       ^                                                              |
       |                                                              |
      (B)                                                             |
+------------+          Client Identifier      +---------------+      |
|           ------(A)-- & Redirection URI ---->+               |      |
|   User-    |                                 | Authorization |      |
|   Agent   ------(B)-- User authenticates --->+     Server    |      |
|            |                                 |               |      |
|           ------(C)-- Authorization Code ---<+               |      |
+---+----+---+                                 +---+------+----+      |
    |    |                                         ^      v           |
   (A)  (C)                                        |      |           |
    |    |                                         |      |           |
    ^    v                                         |      |           |
+---+----+---+                                     |      |           |
|            |>-+(D)-- Authorization Code ---------'      |           |
|    WoT     |         & Redirection URI                  |           |
|  Consumer  |                                            |           |
|            |<-+(E)----- Access Token -------------------'           |
+-----+------+      (w/ Optional Refresh Token)                       |
      v                                                               |
      |                                                               |
      +-----------(F)----- Access WoT --------------------------------+
                           Affordance 

```

It is pretty clear that the device (WoT consumer) needs the authorization endpoint to redirect the User-Agent (and the resource owner) to the right service, but also it needs the token endpoint to satisfy points D and E.

farshidtz:
Thanks for the explanation. For me, TDD is running on the device, and is the software entity which the user agent directly interacts with.

relu91:
> For me, TDD is running on the device, and is the software entity which the user agent directly interacts with.

I think that my comment still applies. They are still logically different entities... however today's we didn't get the chance to really talk about this. Since you are creating a new issue, could you please expand a little better the use case? maybe using a diagram similar to the one above? Thanks!

mmccool:
I've been thinking about this and probably a lot of the "mandatory" items in flows should be optional, if they are provided as part of the protocol.

farshidtz:
@relu91, I've been thinking about this and it is now clear to me that the consumer of TD is NOT a user. So the endpoint are indeed necessary.

Use cases:
- Code flow: A device with rich I/O capabilities is trying to access a central TDD.<br>The user is redirected to Auth Server to authenticate, and redirected back. The UI exchanges the code to get a token from Auth Server (ideally in the backend). The UI can now access TDD with that token. **TD is used by UI.**
- Client flow: A headless device or server application is trying to access a central TDD.<br>The device has its own credentials. It uses them to get a token from Auth Server. The device can now access TDD with that token. **TD is used by device.**
- Device flow: A device with limited UI is trying to access a central TDD.<br>Device uses device flow to get a token from Auth Server. The user is instructed to follow some steps on another computer. Once successful, the device gets a token and can access TDD with it. **TD is used by device.**

Initially, I was thinking about a faulty use case where TDD and WoT Consumer were merged (e.g. gateway device with internal TDD) or where TDD, WoT Consumer, User Agent were merged (e.g. gateway device with internal TDD and rich interface). In this case, not only the human user doesn't need those auth server endpoints, but neither the TD. Perhaps the TD is irrelevant here.

I adapted the [OAuth2 diagrams](https://github.com/w3c/wot-usecases/blob/master/USE-CASES/oauth.md) for TDD. I changed "Access WoT Affordance" labels to "Access Token" since these are referring to what information is passed as part of OAuth2 flows.

### Code
```
  +----------+                                   +-----------+
  | Resource |                                   |    TD     |
  |   Owner  |                                   | Directory +<-------+
  |          |                                   |   (TDD)   |        |
  +----+-----+                                   +-----------+        |
       ^                                                              |
       | (B)                                                          |
       |                                                              |
+------+-----+          Client Identifier      +---------------+      |
|            +------(A) & Redirection URI +--->+               |      |
|   User-    |                                 | Authorization |      |
|   Agent    +------(B) User authenticates +-->+     Server    |      |
|            |                                 |               |      |
|            +<-----(C) Authorization Code +---+               |      |
+---+----+---+                                 +---+------+----+      |
    ^    |                                         ^      v           |
(A) |    | (C)                                     |      |           |
    |    |                                         |      |           |
    |    v                                         |      |           |
+---+----+---+                                     +      |           |
|            +>----(D) Authorization Code +-------+'      |           |
|    WoT     |         & Redirection URI                  |           |
|  Consumer  |                                            +           |
|            +<----(E) Access Token +--------------------+'           |
+-----+------+        (w/ Optional Refresh Token)                     |
      v                                                               |
      |                                                               |
      +------------(F) Access Token ----------------------------------+
```

### Client
```
+-----------+
|    TD     |
| Directory |
|   (TDD)   |
+-----+-----+
      ^
      | (C) Access Token
      |                                       +---------------+
+-----+----+                                  |               |
|          +>---(A) Client Authentication +-->+ Authorization |
|   WoT    |                                  |     Server    |
| Consumer +<---(B) Access Token +-----------<+               |
|          |                                  |               |
+----------+                                  +---------------+
```

### Device
```

+-----------+
|    TD     |
| Directory |
|  (TDD)    |
+----+------+
     ^
     | (G) Access Token
     |        
+----+-----+                                +----------------+
|          +----- (A) Client Identifier +-->+                |
|          |                                |                |
|          +<---- (B) Device Code,      +--<+                |
|          |          User Code,            |                |
|   WoT    |          & Verification URI    |                |
| Consumer |                                |                |
|          |  [polling]                     |                |
|          +>---+ (E) Device Code       +-->+                |
|          |          & Client Identifier   |                |
|          |                                |  Authorization |
|          +<---+ (F) Access Token      +--<+     Server     |
+-----+----+   (& Optional Refresh Token)   |                |
      |                                     |                |
      |                                     |                |
      | (C) User Code & Verification URI    |                |
      |                                     |                |
      |                                     |                |
+-----v----+                                |                |
| End User |                                |                |
|    at    +<---+ (D) End user reviews  +-->+                |
|  Browser |          authorization request |                |
+----------+                                +----------------+

```

relu91:
I am glad that we reached a consensus! and thanks for expanding my point with actual examples 😃 . 

> Initially, I was thinking about a faulty use case where TDD and WoT Consumer were merged (e.g. gateway device with internal TDD) or where TDD, WoT Consumer, User Agent were merged (e.g. gateway device with internal TDD and rich interface). In this case, not only the human user doesn't need those auth server endpoints, but neither the TD. Perhaps the TD is irrelevant here.

yeah, if the TDD and WoT Consumer are practically the same application (sharing even some code) and don't think we need a TD here. However, if they are designed as microservices  TD might still have a place. Probably the only flow that makes sense in this use-case is `client`. 

> I adapted the OAuth2 diagrams for TDD. I changed "Access WoT Affordance" labels to "Access Token" since these are referring to what information is passed as part of OAuth2 flows.

Well, technically the token is exchanged together within the HTTP request to access the WoT affordance, but I agree that it might be confusing. What about: "Access WoT affordance with the token"?

farshidtz:
> > I adapted the OAuth2 diagrams for TDD. I changed "Access WoT Affordance" labels to "Access Token" since these are referring to what information is passed as part of OAuth2 flows.
> 
> Well, technically the token is exchanged together within the HTTP request to access the WoT affordance, but I agree that it might be confusing. What about: "Access WoT affordance with the token"?

I set as only Access Token because on some other arrows, only what is passed is in label (e.g. Client Identifier, Device Code). Also wanted to build my case for this:

It is worth noting this whole definition must be given while TDD only cares about receiving an Access Token (regardless of how):
```
   "securityDefinitions": {
        "oauth2_code": {
            "scheme": "oauth2",
            "flow": "code",
            "authorization": "https://auth.example.com/authorization",
            "token": "https://auth.example.com/token",
            "scopes": [
                "write",
                "read",
                "search"
            ]
        },
        "oauth2_client": {
            "scheme": "oauth2",
            "flow": "client",
            "token": "https://auth.example.com/token",
            "scopes": [
                "write",
                "read",
                "search"
            ]
        },
        "oauth2_device": {
            "scheme": "oauth2",
            "flow": "device",
            "authorization": "https://auth.example.com/device_authorization",
            "token": "https://auth.example.com/token",
            "scopes": [
                "write",
                "read",
                "search"
            ]
        },
        "combination": {
            "scheme": "combination",
            "oneOf": [
                "oauth2_code",
                "oauth2_client",
                "oauth2_device"
            ]
        }
    },
    "security": "combination",
```

Might be worth looking at an [alternative](https://github.com/w3c/wot-thing-description/issues/929):
```
   "securityDefinitions": {
        "oauth2_sc": {
            "scheme": "oauth2",
            "flows": ["code", "client", "device"]
            "authorization": "https://auth.example.com/authorization",
            "token": "https://auth.example.com/token",
            "deviceAuthorization": "https://auth.example.com/device_authorization",
            "scopes": [
                "write",
                "read",
                "search"
            ]
        }
    },
    "security": "oauth2_sc",
```

farshidtz:
Thanks @relu91 for review.

**This can now be merged.** 

I added an issue for the describing the missing vocab: https://github.com/w3c/wot-discovery/issues/54

relu91:
```js
   "securityDefinitions": {
        "oauth2_sc": {
            "scheme": "oauth2",
            "flows": ["code", "client", "device"]
            "authorization": "https://auth.example.com/authorization",
            "token": "https://auth.example.com/token",
            "deviceAuthorization": "https://auth.example.com/device_authorization",
            "scopes": [
                "write",
                "read",
                "search"
            ]
        }
    },
    "security": "oauth2_sc",
```
super compact! however IMHO a little bit messy; I'd prefer the current situation...

mmccool:
We decided to defer merging this until one of you can make the Discovery meeting.  But it's looking good.  To clarify one point: scopes in forms are about what scope is needed to access a form.  Scopes in the scheme are about what scopes can be provided by the authorization server.  My understanding is that scopes are just strings that get encoded into tokens by the auth server, then are checked when used to access some resource.   There must be at least one match between the scopes listed in the token and the scopes required by a particular form.

One slightly weird thing is that you might have different scopes on different forms in the same interaction.   I guess this is ok, you might use different URLs for different scopes (for instance).   It does mean the consumer might have to be a little careful about using the form the consumer actually has authorizations for.

directory.td.json
```diff
@@ -16,9 +16,38 @@
...
+            "authorization": "https://auth.example.com/device_authorization",
+            "token": "https://auth.example.com/token",
+            "scopes": [
+                "write",
+                "read",
+                "search"
+            ]
+        },
+        "combination": {
+            "scheme": "combination",
```
mmccool, line 42: In general, better to name schemes something else like "combo_sc", etc. to avoid confusion.

Issue #42, opened 2020-08-03
Decide how to securely do notifications from a Directory


mmccool:
If we do native HTTP approaches, there might be a problem with sleeping devices.  Polling is not efficient.
So a scheme like MQTT would make more sense, but we need to use a secure variant.
We could also support a scheme where the device could explicitly ask for updates (eg get a history of notifications it might have missed).   The devices that don't use an HTTP-native notification would be expected to check in.

mmccool:
Needs to work through firewalls.  SSEs do.  Webhooks not.  Conexxus uses SSE for this use case, as well as websockets (using SSE streams).   So SSE seems like the best option.   But... AWS does not support SSE in lambdas.  The SSE over websockets is interesting but not currently a supported subprotocol.

mmccool:
Conclusion: let's use SSE.

farshidtz:
Relevant issue to discuss the functional requirements: https://github.com/w3c/wot-discovery/issues/28

farshidtz:
Added a draft PR to describe the interaction on top of SSE.

One technical limitation is that early SSE implementations (including HTML5 [EventSource](https://www.w3schools.com/html/html5_serversentevents.asp)) do not allow setting custom headers in the initial HTTP request. Authorization header is required for Oauth2 Client Credentials flow and passing it in query is not advised:

https://tools.ietf.org/html/rfc6750#section-2.3
> Because of the security weaknesses associated with the URI method
   (see Section 5), including the high likelihood that the URL
   containing the access token will be logged, it SHOULD NOT be used
   unless it is impossible to transport the access token in the
   "Authorization" request header field or the HTTP request entity-body.
   Resource servers MAY support this method.

There are of course polyfills for browsers and modern libraries which allow setting Authorization header.

Issue #28, opened 2020-06-22
Directories should notify subscribers when a TD is updated


mmccool:
Directories should support a capability to notify a subscriber if a TD changes.
Such changes can include rotation of IDs in order to discourage tracking and enhance privacy (only authorized subscribers would get the new ID).

farshidtz:
There should also be a mechanism to keep this information for disconnected or late subscribers. For example, MQTT has persistent sessions and retained messages concepts, but relying on a broker adds additional complexity such as the broker access control.

A request-response alternative could be to redirect **authorized users** from the old resource location to the new one (e.g. HTTP 301 redirect). 

Off topic: Need to define a RESTful interaction for changing the ID.

relu91:
I just want to mention that directories might have also the capability to notify subscribers if something related to one query changes. For example, discovery clients might subscribe not to a set of prefixed events (i.e. `TDAdded`, `TDRemoved`), but using a full-blown discovery query like the following: 
```sparql
select ?td where {
      ?td rdf:type wot:Thing;
            rdf:type  sosa:Sensor;
     .....
}
```` 
Consequently, the directory might also implement a query-based subscription protocol.

So we could still need this notification mechanism even if we end up with static TDs.

We had some experiments last year. You can find a toy implementation of a Thing Directory capable of handling this notification mechanism [here](https://github.com/relu91/SEPAThingDir) (it is quite outdated but I can revamp a little bit before the next meeting) and, as a reference, it was used as a Thing Directory service inside of our WoT Store implementation (See [here](https://w3c.github.io/wot/workshop/ws2/Papers/14%20-%20Sciullo%20et%20al,%20Univ.%20Bologna%20-%20WoT%20Store.pdf)).

zolkis:
IMHO some directories might support TD updates, others may not need to.
But TDs do support versioning - that could be taken as the basis for implementing an update mechanism, optional or not.
If a newer version of a TD is available, there could be an event for that, if the directory declares support for it.
That event could be standardized (for name and content), being supported _implicitly_, or left to the TD to define it _explicitly_..

I am leaning towards letting directories be more like _applications_, described by TDs, as it leaves most room for versatility.
Except when we are coming from security side, they might also be like _app stores_, imposing constraints, e.g. serving vetted TDs in a secure way to authorized clients - which should be the main focus in standardizing directories IMHO.

benfrancis:
> If a newer version of a TD is available, there could be an event for that

[...]

PR #51, opened 2020-08-15
Add description of the notification API

farshidtz:
The event API based on SSE, as discussed in #42.

- I am defining only one event affordance called `registration` for three event types (createdTD, updatedTD, deletedTD) because this enables subscription to a single or multiple events in a single SSE stream. Alternatively, we can separate them to three event affordances and use multiple SSE streams; when using HTTP/2, the streams could get multiplexed over the same TCP connection.

- Both `id` and `types` are optional; without them the authorized user can subscribe to all events on all TDs.

- Other query parameters may be supported by passing query expressions (e.g. jsonpath, sparql), if required per [this comment](https://github.com/w3c/wot-discovery/issues/28#issuecomment-647683425).

TODO:
- [x] Add events API to TD.
- [x] Add assertion: The event type must be used by the server to deliver only requested events, regardless of SSE's native client-side event filtering.
- [x] Add assertion: The server should keep a history of events and deliver them to re-connecting clients (based on the provided SSE's `lastEventId`) -> included in SSE specs.
- [x] Change types to type
- [x] Describe the format of SSE `data`.
- [x] Describe the use of `Authorization` header. -> added editorial note as this may go to the security section.
- [x] Change TD `id` in query to something else to avoid confusion with event ID.
- [x] Change description of TD ID to "identifier of Thing Description in directory"

***
<a href="https://pr-preview.s3.amazonaws.com/farshidtz/wot-discovery/pull/51.html" title="Last updated on Sep 7, 2020, 1:23 PM UTC (26c7360)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-discovery/51/d964f85...farshidtz:26c7360.html" title="Last updated on Sep 7, 2020, 1:23 PM UTC (26c7360)">Diff</a>

### DIFF
--- a/directory.td.json
+++ b/directory.td.json
@@ -154,10 +154,6 @@
         "registration": {
             "description": "TD registration events",
             "uriVariables": {
-                "td_id": {
-                    "title": "Identifier of TD in directory",
-                    "type": "string"
-                },
                 "type": {
                     "title": "Event type",
                     "type": "string",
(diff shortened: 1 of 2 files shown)


mmccool:
type/types, also would be nice to have text in the Notification section.

mmccool:
I think making diffs be returned iff there is a query parameter.   I think providing diffs can be optional, but if it is provided it needs to be specified in the TD for the directory instance.  During 2020.08.31 Cristiano expressed the same opinion, in favor of making the feature optional.    In general I think only essential functionality should be mandatory.

mmccool:
Also, may want to change description of "id" to "identifier of Thing Description in directory"; this is correct no matter whether we decide to make this the same as the TD's id field or not (and also covers the case when the TD does not have an id field and the directory needs to make one up).

Issue #53, opened 2020-08-17
Clarifying Introduction Mechanisms


k-toumura:
To clarify introduction mechanisms, I wrote a diagram of discovery flow.

![discovery-lifecycle](https://user-images.githubusercontent.com/30313213/90398882-0c76f580-e0d5-11ea-81d6-3538c710abf5.png)

- Are there any other mechanisms we should add?
- Basically, introduction mechanism resolves to the URL of Thing Description of Thing or Directory.   Should we include a mechanism that resolves directly to the Thing Description (e.g. DID resolves to a DID document, which may contain a TD)?

farshidtz:
Related to DNS-SD for service type and TXT record definition:
- https://github.com/w3c/wot-discovery/issues/5
- https://github.com/w3c/wot-testing/issues/34
- [Fujitsu use case](https://github.com/w3c/wot-discovery/tree/master/prior-work/fujitsu)

k-toumura:
Use of [RFC8615 Well-Known Uniform Resource Identifiers (URIs)](https://tools.ietf.org/html/rfc8615)
- TD self-hosting on well-known URI, such as `/.well-known/wot-thing-description`
- similar to CoRE devices ([RFC6690 Constrained RESTful Environments (CoRE) Link Format](https://tools.ietf.org/html/rfc6690))

benfrancis:
See [mDNS/DNS-SD discovery mechanism](https://iot.mozilla.org/wot/#mdns-dns-sd) used by Mozilla's WebThings Gateway and WebThings Framework libraries. Discussion on naming choices [here](https://github.com/mozilla-iot/gateway/issues/696).

danielpeintner:
>     * TD self-hosting on well-known URI, such as `/.well-known/wot-thing-description`

Just for your information: In node-wot we decided to report the TD when retrieving the thing URL itself.
Having said that, I am fine with reporting  the TD under any well known URI ... it just needs to be pre-scribed *what* the URI should be for a given protocol

k-toumura:
Revise the diagram based on above discussion:
![discovery-lifecycle](https://user-images.githubusercontent.com/30313213/93015860-d3f50980-f5f7-11ea-81aa-a271d3fad7cf.png)

mmccool:
[...]

PR #60, opened 2020-09-06
First draft of introduction mechanisms section

k-toumura:
Add "Section 5. Introduction Mechanisms" based on discussion in issue #53.

***
<a href="https://pr-preview.s3.amazonaws.com/k-toumura/wot-discovery/pull/60.html" title="Last updated on Sep 14, 2020, 11:41 AM UTC (25d483c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-discovery/60/d8efcf7...k-toumura:25d483c.html" title="Last updated on Sep 14, 2020, 11:41 AM UTC (25d483c)">Diff</a>


index.html
```diff
@@ -251,19 +265,108 @@ <h2>Direct</h2>
...
+                This can be also be used to discover them on the same link by combining Multicast DNS (mDNS)[[RFC6762]].
+            </p>
+            <p>
+                In DNS-SD, format of the Service Instance Name is <code>Instance.Service.Domain</code>. 
+                The Service part is a pair of labels following the conventions of [[RFC2782]].
+                The first label has an underscore followed by the Service Name,
+                and the second label describes the protocol.
+            </p>
+            <p>
+                The Service Name to indicate the Thing MUST be <code>_thing._sub._wot</code>.
```
farshidtz, line 297: I think `_thing` is too abstract as a subtype for physical devices; directory is also a Thing. I suggest using `_device` and `_directory` subtypes instead.

index.html
```diff
@@ -251,19 +265,108 @@ <h2>Direct</h2>
...
+                And the Service Name to indicate the Directory Service MUST be <code>_directory._sub._wot</code>.
```
farshidtz, line 298: I think `_thing` is too abstract as a subtype for physical devices; directory is also a Thing. I suggest using `_device` and `_directory` subtypes instead.

farshidtz: Also worth noting that the use of subtypes makes it possible to browse devices and directories separately or at the same time. However, when browsed at the same time, it won't be possible to tell them apart because subtype value won't appear in the results. This was discussed in https://github.com/w3c/wot-testing/issues/34. 

We can assume that all registrations are for devices and have an additional TXT record to distinguish directories.

mmccool: Really these should describe types of "exploration" services.   A directory is clearly one type, so we should have "directory" as one.    The other kind of exploration service though is when a Thing self-describes and hosts its own TD, so the type could be "self".   The other option is to leave off the subtype, but we should not imply that all things can self-describe.   It also needs to work with URL-only introductions... which should probably always point at a "self" endpoint.

mmccool:
Proposal: just use _wot for everything, and _directory._sub._wot for directory APIs.   A _wot link type is also implied if only a URL is given (needed for some intro mechanisms).   In both cases the link points at a TD.  The only difference is that the user does not need to read the TD and look for a "Directory" in @type to know if it is a a directory; the type tells the user that is it.

k-toumura:
Updated the document according to above discussion.
And Also updated a description of DID-based discovery.

index.html
```diff
@@ -251,20 +265,133 @@ <h2>Direct</h2>
...
+                The Service part is a pair of labels following the conventions of [[RFC2782]].
+                The first label has an underscore followed by the Service Name,
+                and the second label describes the protocol.
+            </p>
+            <p>
+                The Service Name to indicate the Thing or Directory Service MUST be <code>_wot</code>.
+                And the Service Name to indicate the Directory Service MUST be <code>_directory._sub._wot</code>.
             </p>
+            <p class="ednote" title="Service Names in existing implementations">
+                The Service Names <code>_thing._sub._wot</code> and <code>_directory._sub._wot</code>
```
farshidtz, line 301: Thanks for the changes. 

Just a minor issue in editorial note: `_thing._sub._wot` should be change to `_wot`. Probably left unchanged from the previous version.

k-toumura:
Fix a Service Name in Editor's Note (`_thing._sub._wot` -> `_wot`).

[Rendered version](https://cdn.statically.io/gh/w3c/wot-discovery/25d483c04df8394ca903fa3f4bd2801fb48ccf22/index.html).

Issue #97, opened 2020-11-16
How do we handle federation in SPARQL queries


mmccool:
SPARQL federation has the following issues:
1. Potential for DoS attack vector, since SPARQL queries are expensive and federation amplifies this
2. SPARQL federation requires that the URLs of other SPARQL endpoints are known in advance
3. If one directory links to another, and the client follows the link, is that considered federation?   We need to read the spec carefully; service endpoints vs. documents.   Is SPARQL definition appropriate or do we need our own (more constrained) definition?

mmccool:
My comment on 2: we may get multiple results from an "Introduction" query, and perhaps "federation" could be over these results.

mmccool:
Proposal:
1. (Limited to delegated/recursive federation, where a directory queries other directories on a client's behalf).  SPARQL does not allow "recursion" per se, delegated endpoints must be known in advance.   There are also other ways to limit DoS attacks, eg. rate-limiting).  Allow SPARQL delegation in a directory implementation, but do not require it, and add an editor's note regarding the DoS concern.
2. A client may do an introduction and get multiple results, send queries to exploration services (eg directories and/or peer-to-peer) in parallel, and merge the results.
3. Returning links to other directories so a client can also query them is permitted.  See #34 

Not included:
* SPARQL implementations using delegation/federation internally are not included in 1.   We are only constraining delegation/federation between WoT TDDs.

To do: create text for 1 and 2.  3 will be handled during resolution of #34

mmccool:
So... PR has a SHOULD assertion.  For my proposal above (allow it, but do not require it) I was thinking this should be a MAY.  SHOULD means it is recommended., which is something we should discuss.  Certainly if it is implemented, it must follow the SPARQL spec, and we seem to agree that it should be optional.  So the thing we have to decide is "how" optional it is.  We should also look very carefully at [RFC2119](https://tools.ietf.org/html/rfc2119), the definitions of MAY and SHOULD are specific: 

SHOULD:  This word, or the adjective "RECOMMENDED", mean that there
   may exist valid reasons in particular circumstances to ignore a
   particular item, but the full implications must be understood and
   carefully weighed before choosing a different course.

MAY:  This word, or the adjective "OPTIONAL", mean that an item is
   truly optional.  One vendor may choose to include the item because a
   particular marketplace requires it or because the vendor feels that
   it enhances the product while another vendor may omit the same item.
   An implementation which does not include a particular option MUST be
[...]

PR #105, opened 2020-12-07
Updated document with federation

AndreaCimminoArriaga:
Text regarding federation included, as discussed in #97

resolves https://github.com/w3c/wot-discovery/issues/97

***
<a href="https://pr-preview.s3.amazonaws.com/AndreaCimminoArriaga/wot-discovery/pull/105.html" title="Last updated on Dec 14, 2020, 1:10 PM UTC (81c9c60)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-discovery/105/b0ee195...AndreaCimminoArriaga:81c9c60.html" title="Last updated on Dec 14, 2020, 1:10 PM UTC (81c9c60)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -983,7 +983,8 @@ img.wot-diagram {
                 <li>501 (Not Implemented): SPARQL API not supported.</li>
               </ul>
             </p>
-	<span class="rfc2119-assertion" id="tdd-distributed-search-semantic">if implemented, the SPARQL API SHOULD implement the SPARQL 1.1 Federated Query standard [[sparql11-overview]]</span>
+	<span class="rfc2119-assertion" id="tdd-distributed-search-semantic">A WoT Thing Description Directory MAY implement federation in its SPARQL query API.</span>
+	<span class="rfc2119-assertion" id="tdd-distributed-search-semantic-imp">If implemented, the SPARQL API MUST implement the SPARQL 1.1 Federated Query standard [[sparql11-overview]].</span>
 	<p class="ednote" title="SSE Authorization Header">
              The WoT Thing Directory SHOULD implement mechanisms to prevent DDoS attacks in the SPARQL search API
         </p>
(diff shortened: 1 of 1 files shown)


mmccool (changes requested):
Don't think this quite captures what is discussed in the issue.  In particular I would have expected a MAY rather than a SHOULD assertion.  Let's talk about it a bit more in the  next discovery call.  I added a few comments to the issue.

mmccool:
Also, "if implemented" implies MAY, but what the assertion actually says is following the SPARQL spec is optional, and I think that should be a MUST.

index.html
```diff
@@ -983,7 +983,11 @@ <h4>Semantic search: SPARQL</h4>
                 <li>501 (Not Implemented): SPARQL API not supported.</li>
               </ul>
             </p>
-      			</section>
+	<span class="rfc2119-assertion" id="tdd-distributed-search-semantic">if implemented, the SPARQL API SHOULD implement the SPARQL 1.1 Federated Query standard [[sparql11-overview]]</span>
+	<p class="ednote" title="SSE Authorization Header">
+             The WoT Thing Directory SHOULD implement mechanisms to prevent DDoS attacks in the SPARQL search API
```
mmccool, line 988: Change to use two assertions:
1. A WoT Thing Description Directory MAY implement federation in its SPARQL query API.
2. If implemented, the SPARQL API MUST implement the SPARQL 1.1 Federated Query standard [[sparql11-overview]]

Also, use WoT Thing Description Directory (TDD to avoid confusion with "TD").

Resolved in the meeting 07.12.2020 to merge after these changes.

AndreaCimminoArriaga: Changes requested addressed

mmccool: Updates look good, but ed note is formatted as an assertion and has some other issues (title is wrong, for instance).  I'm going to merge and then edit the note...

mmccool (approved):

Issue #98, opened 2020-11-16
Need information model for directory


mmccool:
We need an overall information model for all the data in a directory.  Does it include information besides TDs?   Eg out-of-band information (TTL, update timestamps, location...), links (eg to other directories), Thing Models, etc.  
Can SPARQL queries be over the entire model?

Related issues:
* Federation: https://github.com/w3c/wot-discovery/issues/97
* Out-of-band info: https://github.com/w3c/wot-discovery/issues/30
* Links: https://github.com/w3c/wot-discovery/issues/34
* TTL: https://github.com/w3c/wot-discovery/issues/18
* Geolocation: https://github.com/w3c/wot-discovery/labels/Geolocation
* Timestamps: https://github.com/w3c/wot-discovery/labels/Timestamps

mmccool:
To be clearer, I'm thinking each TD record would "internally" look like this:

<td_record> :=
{
    // metadata
   "ttl": <ttl>,
   "location": <location>,
   "lastUpdated": <timestamp>,
   ...
    // tds (not modified by directory)
    "td": <parsed TD JSON>,
    "original": <raw string, for signing preservation purposes>
}

and the directory would basically stored as a JSON-LD object containing a set of these:

<directory> := { 
  "@context": <directory ontology>,
   <local_id>: <td_record>, ...
}

See however discussion in https://github.com/w3c/wot-discovery/issues/18, which already discussed something similar to this (and the issues with it...)

farshidtz:
I provided a summary of all proposals in my [presentation](https://github.com/w3c/wot/blob/master/PRESENTATIONS/2020-06-online-f2f/2020-06-22-Discovery-ThingDirectory-Tavakolizadeh.pdf) (slides 9-11) back in June 2020 F2F. But an up-to-date summary would be more useful.

For the API spec, it is important to know how the directory **exposes** the resources, and that is what I will cover below. 

I will only include `created` and `modified` as out-of-band attributes but there could be [others](https://github.com/w3c/wot-discovery/issues/30). I will particularly leave out `ttl` and `expires` because we could rely on [protocol-specific mechanisms](https://github.com/w3c/wot-thing-description/issues/916#issuecomment-645870661) for those. In such case, it is safe to assume that all out-of-band information are system-generated and read-only.

## 1. Wrapped TD
/td-wrapper/urn:example:1234
```json
{
  "@context":"https://www.w3.org/2019/wot/tdd/v1",
[...]

Issue #54, opened 2020-08-18
Need to define the discovery types and vocabulary


farshidtz:
Some types and vocabulary used in Discovery spec and [Directory's TD](https://github.com/w3c/wot-discovery/blob/master/directory.td.json) are not part of the [TD context](https://github.com/w3c/wot-thing-description/blob/master/context/td-context-1.1.jsonld). 

We can extend the context locally (see [context-extensions](https://www.w3.org/TR/wot-thing-description/#sec-context-extensions)) or have them added to TD's main context.

To be included:
- [x] `combo` scheme in SecurityScheme (discussed in https://github.com/w3c/wot-thing-description/issues/901) - added to [TD 1.1](https://github.com/w3c/wot-discovery/issues/54)
- [x] `additionalResponses` in Form (discussed in https://github.com/w3c/wot-thing-description/issues/617) - added to [TD 1.1](https://github.com/w3c/wot-thing-description/pull/1042)
- [x] type for directory TDs (discussed in https://github.com/w3c/wot-discovery/issues/43) - added to [Discovery context](https://github.com/w3c/wot-discovery/blob/master/context/discovery-context.jsonld)
- [x] type for link TDs (see https://github.com/w3c/wot-discovery/issues/34#issuecomment-740009713) - added to [Discovery context](https://github.com/w3c/wot-discovery/blob/master/context/discovery-context.jsonld)
- [x] Registration annotations (see https://github.com/w3c/wot-discovery/issues/98#issuecomment-739245916) - added to [Discovery context](https://github.com/w3c/wot-discovery/blob/master/context/discovery-context.jsonld)

mmccool:
Should create issue(s) in the wot-thing-description for these so the TD task force sees them.   Combination scheme is still under discussion; the two need issues.   Please link here.

farshidtz:
alternateResponses is being discussed in https://github.com/w3c/wot-thing-description/issues/617

relu91:
See my comment: https://github.com/w3c/wot-discovery/issues/43#issuecomment-762186016

mmccool:
I think the last one (registration annotation) should probably be in its own context, since it only appears in directory responses when requested.  See also comment from @relu91

mmccool:
Note also "combo" scheme has been closed.

farshidtz:
[...]

Issue #30, opened 2020-06-22
Define Out-of-Band Information


mmccool:
Assuming directories cannot modify the TDs they store, what out-of-band information should a directory associate with each TD in its information model?   My thought is that this information should be subject to query as well, so TDs are just part of each "record" in the directory's database.  Here are some candidates:
- local id.   In case the TD does not contain a global id, the database should be assigning a local id to be used to retrieve a specific TD.
- timestamps.  Date of first creation, date of last update
- ttl.  how long before the TD is automatically deleted.

benfrancis:
I would suggest any metadata like this should be stored in the TD itself, or via the protocol the TD is served by (e.g. Expired and Last-Modified headers of HTTP). 

I wouldn't think the internal database IDs of Thing Descriptions should be exposed to WoT clients by directories, the Thing Description URI (or ID member) should be sufficient.

All a directory really needs to serve is a list of URLs from which the Thing Descriptions can be retrieved, or the content of the Thing Descriptions themselves.

farshidtz:
Some things have been discussed in https://github.com/w3c/wot-discovery/issues/18.

mmccool:
@benfrancis we do have some specific issues where adding information to the TD is troublesome:
1. Signed TDs.  Sure, we can use proof chaining, but then the consumer has to trust the directory as well as the endpoint.  If we just want to add metadata to a TD (like the TTL of the TD *in this particular directory*) then it seems like we don't want to break signing for that.
2. Encrypted TDs.  It's possible a directory might want to store and transmit a TD upon request that it can't itself view.  Of course in this case the consumer would need to know the key used to look it up (during registration the registering entity would have to provide a "key", which may or may not be the same as the ID inside the TD, which the directory can't access...).   BTW by key I don't mean the decryption key, I just mean some value used to find the encrypted TD in the directory.
3. Geolocation.  I've been thinking a lot about how to deal with geolocation queries.  This information can be encoded in the TD (for static location, e.g. infrastructure devices) or can be provided by a property (for dynamic location, e.g. mobile devices).  It might also be inferred by the directory (e.g. within WiFi range).   So we could store the directories best guess as the location of the Thing in metadata and use that for queries.   Then all the geolocation is in one unified place, making setting up queries easier.

mmccool:
Anyhow, another issue is that IDs are optional in TDs.  So yeah, the internal ID would be a useful key except we need to deal with TDs that omit them.

benfrancis:
[...]

PR #112, opened 2021-01-24
Describe the information model

### LINKED ISSUE #34: Add links section to directory information model
See also https://github.com/w3c/wot-discovery/issues/32

It has been suggested that in some cases we want to store links to TDs or to other directories in a directory rather than just TDs.  We have two possible ways to do this:
1. Store TDs with just links (note the interactions, etc. in a TD are optional, so...) 
2. Have a separate links section in the information model of the directory.

Both are actually useful.  Option 1 can be used if searchable metadata, eg. semantic annotation, is needed (eg for a remote directory where we want to list the types of Things indexed by that directory).  But option 2 is also simple and useful, and can do things like have concrete links to all TDs stored on the directory, as well as links to self-identifying devices that for some reason (eg private information in the TD) don't want to store a copy of their TD in the directory.   Note that links may be to both other information stored in the directory itself and to remote information (eg other directories).

The links section should probably have some metadata, i.e. at least a relation type (probably the same as the relation types we need to define for DID documents, and able at least to distinguish things from directories), and maybe also an optional spatial location or region (to facilitate federated  spatial queries over a distributed directory service).   But any more than that, and a TD should be used.

farshidtz:
- [x] Conceptual class diagram (related to #98)
- [x] Directory type (related to #54)
- [x] Link type and example (closes #34)
- [x] Discovery context (related to #54)
- [x] Example for tdd:registration extension (related to #98 and #30)
- [x] ~~API specification and description for TD enrichment with registration info~~ Added ednote, will cover in another PR.

The preview doesn't show the figure. I'll add it here (more recent diagrams below):

***
<a href="https://pr-preview.s3.amazonaws.com/farshidtz/wot-discovery/pull/112.html" title="Last updated on Feb 1, 2021, 4:35 PM UTC (5c2480d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-discovery/112/52a3a5f...farshidtz:5c2480d.html" title="Last updated on Feb 1, 2021, 4:35 PM UTC (5c2480d)">Diff</a>


AndreaCimminoArriaga:
I like the model, however I would change the types names for others since right now they seem a bit confusing to me. Maybe Directory, Thing Description, Directory Description, and Link Description ?

farshidtz:
> I like the model, however I would change the types names for others since right now they seem a bit confusing to me. Maybe Directory, Thing Description, Directory Description, and Link Description ?

I like the suggested naming. After some discussion with @AndreaCimminoArriaga, here is a new diagram:

farshidtz:
Added Thing for self-described devices; the model is also relevant to those.

AndreaCimminoArriaga (approved):

farshidtz:
After a discussion with @AndreaCimminoArriaga, updated the type in directory API's TD and added annotations and arrowheads to diagram.

Issue #67, opened 2020-09-07
Review current directory API proposal for privacy and security


mmccool:
The current directory API proposal should be reviewed for security and privacy risks

farshidtz:
Duplicated by:
- #254

PR #113, opened 2021-02-03
Security and Privacy Considerations

mmccool:
Adding additional content to the Security and Privacy Considerations section.
See issue https://github.com/w3c/wot-discovery/issues/67 and issue https://github.com/w3c/wot-security/issues/196

***
<a href="https://pr-preview.s3.amazonaws.com/mmccool/wot-discovery/pull/113.html" title="Last updated on Feb 22, 2021, 2:13 PM UTC (40285c7)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-discovery/113/a097dbf...mmccool:40285c7.html" title="Last updated on Feb 22, 2021, 2:13 PM UTC (40285c7)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1148,22 +1148,94 @@ img.wot-diagram {
             privacy issues, see the <em>WoT Security and Privacy Guidelines</em>
             specification [[?WOT-SECURITY]].
         </p>
-        <p class="ednote" title="Discovery Security and Privacy Concerns and Mitigations">
-        To do, some discussion of general security and privacy concerns and mitigations.
-        Note that the architecture above is designed to address many such points, for
-        example the two-phase approach and "authorization before metadata release"
-        principles, so this would be 
-        a summary and a recap.
+        <p>
+        The WoT discovery architecture is designed to avoid a dependence on the security
+        and privacy of existing discovery schemes by using a two-phase approach and
+        requiring authorization before metadata release.  However several security and
+        privacy risks still exist.  These are listed below along with possible mitigations.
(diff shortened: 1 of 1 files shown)


mmccool:
Marked as not-WIP however at present I only have two risks: DDoS and location tracking.   Suggestions for other risks welcome.  Certain issues mentioned in the above-linked issues should be dealt with elsewhere in the document.  For example, JSON path injection attacks should be designed out by limiting the features that can be used in JS embedded in JSON path queries.

mmccool:
* Access controls for explicit location data should me mentioned
* Negative location inferencing can be mitigated by having long time to live (e.g. one month in the home use case)
* We should consider things/ask PING what they would likely be concerned about

mmccool:
Comment:
* Peer-to-Peer; change to Interpersonal and Interinstitutional?
* Not normative; there are some comments above and in the issues that should be dealt with separately with normative changes, though.
* I did capture the entire discussion from the issues above and embedded it in a comment for easy reference; this should be stripped in the final version

mmccool:
* add references from S&P section (which is informative) to normative items addressing particular concerns.  This will make it easier for Security/Privacy reviews to find risks and mitigations.  
* We might even mention a risk but just point to a normative mitigation that addresses it (e.g. query injection -> limited expressive power of queries)
* 
For example, for DDoS, we should reference specific mechanisms specified to prevent query injections, use of arbitrary JS in JSON Path, etc. (last point might be dealt with in IETF JSON path spec though, need to check).

PR #114, opened 2021-02-05
WIP: Proposal for geolocation

mmccool:
WIP: not yet complete, do not merge.  Comments welcome.

Proposal for geolocation information model and query filter mechanisms for discovery.   Examples of TD JSON-LD encodings pulled from 2020.09 plugfest.

### DIFF
--- a/proposals/geolocation.md
+++ b/proposals/geolocation.md
@@ -363,6 +363,8 @@ In general, BIMs (Building Information Management systems) define ontologies for
 logical relationships indoors.
 
 ### JSON-LD Examples in TD
+*To do:* Examples are not consistent with definitions above, e.g. use of `x`, `y`, and `z` for
+elements of `velocity`, etc.  Needs to be fixed.
 
 #### Static Location
 Static locations may not change or only be infrequently updated.
--- a/proposals/geolocation.md
+++ b/proposals/geolocation.md
@@ -108,7 +108,7 @@ latitude, and east being positive for longitude.  Units for latitude
 and longitude are always degrees.  Accuracy is also given but as a distance,
 so the area of uncertainty is a cylinder perpendicular to the Earth's surface.
 
-Elements:
+*Elements:*
 
 | Element Name   | Type   |  Min |  Max | Default | Description                                                                  |
 |----------------|--------|------|------|---------|------------------------------------------------------------------------------|
@@ -147,7 +147,7 @@ mean sea level.
 If an elevation is also given, then the depth is relative to that elevation
 instead of the surface.
 
-Elements:
+*Elements:*
 
 | Element Name   | Type   |  Min |  Max | Default | Description                                                                  |
 |----------------|--------|------|------|---------|------------------------------------------------------------------------------|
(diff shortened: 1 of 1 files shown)


proposals/geolocation.md
```diff
@@ -0,0 +1,404 @@
+# WoT Discovery
+## Geolocation
+Geolocation support for discovery includes two loosely-coupled
+specifications: an information model for geolocation data itself,
+which might be encoded in TDs or returned as data from properties,
+and a query mechanism to filter such data by location and time.
```
cperey, line 6: do we want/need to distinguish between the term "position" and "location?

proposals/geolocation.md
```diff
@@ -0,0 +1,404 @@
...
+#### Logical Regions and Relationships
+Sometimes is it more convenient to refer to a labelled region or zone
+than a measured location.  These "logical" locations can be used instead
+of or in addition to numerical locations.
+Example of logical locations include:
+* Named rooms in a building
+* Floor in building
+* Plots in a city
+* Roads
+It may also be useful to represent
```
cperey, line 151: Maybe another important use of OWL ?

proposals/geolocation.md
```diff
@@ -0,0 +1,404 @@
...
+          "contentType": "application/json"
+        }
+      ]
+    }
+  }
+}
+
+```
+
+#### External Geolocation Device
```
cperey, line 353: The Geolocation can also be provided by a service which the Thing will send a request to (e.g., send an image and, based on visual localization service, the position and orientation are returned to the Thing.

proposals/geolocation.md
```diff
@@ -0,0 +1,404 @@
...
+specifications: an information model for geolocation data itself,
+which might be encoded in TDs or returned as data from properties,
+and a query mechanism to filter such data by location and time.
+
+Note: filtering by time of last update potentially could be broken out
+as a separate proposal since it is applicable to data other than
+location.  However, location is special since the Directory needs to
+track it to provide geolocation queries.  Queries over other kinds of
+data would require additional requirements in the Directory e.g. a 
+time-series database, and this is currently out of scope.
```
farshidtz, line 13: If the directory keeps a history of changes to each TD, it could be used to track static location updates.

proposals/geolocation.md
mmccool: probably I should be more careful in my terminlogy and use one or the other.  I'm not sure there is a distinction.  I do think is a little confusing to use "geolocation" for the link and then use "location" as just PART of the total set of geolocation information, however.  Something to discuss with the experts

proposals/geolocation.md
```diff
@@ -0,0 +1,404 @@
...
+together on a map.  This has many applications:
+* City planning 
+* Emergency response
+* Pollution monitoring
+* Sensor placement and orientation during installation
+* Building permits - "call before digging"
+#### AR Guide
+* Situational awareness
+* Tourism
+* Maintenance and Repair
```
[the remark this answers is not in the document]
mmccool: Anything specific I could add here?  This is only meant to be a summary, however.  We should add detailed comments to the Use Cases and Requirements document.

proposals/geolocation.md
```diff
@@ -0,0 +1,404 @@
...
+If an elevation is also given, then the depth is relative to that elevation
+instead of the surface.
+
+#### Orientation
+Both 
+* 2D Heading (azimuth) - angle from north
+* Incline (elevation) - deviation from vertical
+* Tilt (rotation around heading)
+Issues:
+* Singularies e.g. at the pole
```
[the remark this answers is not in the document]
mmccool: right, so for example, "heading" as an angle is not defined at the north and south poles.   However... that's what is used in the Web API, and it's a pretty rare case, so we can special-case it, I think.  Other singularities are 3D rotations specified using Euler angles, and has other problems, like order dependence.  A vector/angle specification avoids these issues (related to quaternions, but vector/angle is probably easier to understand, although a coordinate system then needs to be defined, and THAT will have singularities, i.e. "North" may not be well defined at the poles).

proposals/geolocation.md
```diff
@@ -0,0 +1,404 @@
...
+* Building permits - "call before digging"
+#### AR Guide
+* Situational awareness
+* Tourism
+* Maintenance and Repair
+#### Accessibility
+* Location of accessible services
+* Navigational assistance
+
+### Usage Patterns
```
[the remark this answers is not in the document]
mmccool: Yes, that's reasonable.  I will have to look into it.   Here are some links I found, and I see in particular they are targeting a JSON encoding (which is good): https://www.ogc.org/projects/groups/geoposeswg
What I don't see is whether I can get access to a draft without being an OGC member.

proposals/geolocation.md
mmccool: Oh, that's interesting.  Closely related to edge computing.  However, I think that we should put that under "extensions" for now.

proposals/geolocation.md
mmccool: Yes, and there are ontologies already defined for this that we should track down and reuse.

proposals/geolocation.md
mmccool: True, although I think general history opens up a huge can of problems.  So I think we want to target some very specific use cases here.  Basically, if the location is "stale" then I might want to ignore it or give it a higher uncertainty, or estimate the current location based on the last known position and speed.  So the time of last update is important.  Maintaining an entire history might be useful for more accurate trajectory estimates, but I think this is a relatively minor benefit for a lot of additional requirements, privacy issues around data retention, and so forth.  To discuss...

mmccool:
So, I tracked down the OGC GeoPose spec and tried to make my proposal consistent with it, but there are various issues.  Still working on some things as well.  I think the best we can do is to make it "as consistent as possible".   GeoPose is also still in draft state.

Anyhow, I think I will be able to wrap up the information model draft soon, then look at query specs.

PR #115, opened 2021-02-07
Specify self-described exploration

farshidtz:
This PR specifies the self-description mechanism.

Moreover, it adds the fundamental terminologies similar to those in https://www.w3.org/TR/wot-thing-description11/#terminology.

***
<a href="https://pr-preview.s3.amazonaws.com/farshidtz/wot-discovery/pull/115.html" title="Last updated on Feb 14, 2021, 4:31 PM UTC (0ee7abc)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-discovery/115/b122b36...farshidtz:0ee7abc.html" title="Last updated on Feb 14, 2021, 4:31 PM UTC (0ee7abc)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -512,8 +512,38 @@ img.wot-diagram {
         
         <section id="exploration-self" class="normative">
             <h2>Self-description</h2>
-            <p class="ednote" title="Self-Description Overview">
+            <!-- <p class="ednote" title="Self-Description Overview">
             To do: Describe mechanisms for devices to self-describe, hosting their own TDs.
+            </p> -->
+            Self-describing devices are those which serve a Thing Description
+            and expose it at a URL. The URL may be advertised via one of the
+            [[[#introduction-mech]]] or a [[[#exploration-directory]]]. 
+            
+            <p>
+                <span class="rfc2119-assertion" id="self-retrieve"> 
+                    A self-describing device MUST expose its TD over HTTPS (HTTP Over TLS)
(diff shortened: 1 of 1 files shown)


mmccool:
This is a good start and I think it should be merged, perhaps after marking up the assertions with the right "spans" and giving them unique IDs.  However, the reference to WOT-SECURITY is not normative, and the text does not go as far as to state that security is mandatory, etc.  I see at least that it insists on HTTPS, although this is actually troublesome in a "local" network and we may want to also allow CoAPS (CoAP + DTLS).   I should also say in general we avoid saying "HTTPS", instead (e.g. in the security doc) say "HTTP over TLS".   We might also want to mention that one of the purposes of using TLS, etc. is to hide the request itself from listeners (although a listener can still see the IP request happening, at least without SecIP).

farshidtz:
> perhaps after marking up the assertions with the right "spans" and giving them unique IDs.

The assertions are from the existing spec in chapter 5. I have actually added an ednote to highlight the same issue.

Assertions added in this PR have the correct span and class.

farshidtz:
Removed the reference to security spec. 

Added comment (in code) for additional self-describing protocols. In case we add CoAP, we need to also consider [CoAP's Discovery](https://tools.ietf.org/html/rfc7252#section-7) and its relation with the Introduction mechanisms.

Issue #16, opened 2020-04-20
Handle huge set of Thing Descriptions (pagination, streaming, etc.)


sebastiankb:
**Use case**
A TD Directory manage a huge set of TDs, maybe around 1000-10000 TDs. A client queries the TD directory where about 1000 TDs would match.

**Problem statement**
How are the 1000 TDs responded to the client in a resource-efficiently way? Will this be a huge file where all TDs are encapsulated? Or will the TDs be fragmented into blocks and answered? Or will there be a stream?

**First brainstormings**  
1. TDs maybe encoded such as with [EXI/CBOR](https://github.com/w3c/wot-thing-description/issues/885)
2. Rely on a standard that handles such similar use case (e.g., [Block-Wise Transfers](https://tools.ietf.org/html/rfc7959))
3) How would we do something like a paging through a set of TDs?
4) How do established graph databases deal with such use cases?

danielpeintner:
Yet another possibility would be that a discovery call simply returns a list of links to the actual TDs.

egekorkan:
Another idea would be treating it like search engines where the most relevant results are placed first, like 10 TDs, which _hopefully_ does not make a huge document. The client then goes to other "pages" and looks again. Based on the client's processing capabilities, it can ask for more TDs in the first request, like saying that a shopping website should display 100 items in a page based on user preference.

relu91:
IMHO pagination is the best option. We just need to represent a set of TDs; it could be JSON or JSON-LD or more efficient formats as presented in point 1. 

> Yet another possibility would be that a discovery call simply returns a list of links to the actual TDs.

A more sophisticated approach could be a Level Of Detail method. So that only `partial TDs` are returned. How much 'partial' the TD is, it is specified inside the search request. The problem with returning only links is that clients probably would end up to fetch every single TD out of them and, therefore, creating just networking overhead. Anyway, I still prefer pagination over this method.

> How do established graph databases deal with such use cases?

For databases that support SPARQL (i.e. SPARQL endpoints) pagination is handled with a combination of `OFFSET` `LIMIT` and `ORDER BY` query modifiers. Basically, you order by a criteria and then select a portion of the whole RDF Dataset matched using a starting pointer (`OFFSET`) and a length parameter (`LIMIT`). Since the order will be the same in the next queries you can "navigate" through portions of the result.

mmccool:
1. Let's assume all queries return a set of results.  In our case they could be everything from links to full TDs.
[...]

PR #130, opened 2021-03-06
Improvements to paginated TD retrieval spec

farshidtz:
- [x] Specify query arguments based on the original proposal
- [x] Specify headers and modify to match https://www.w3.org/TR/ldp-paging/ as suggested in https://github.com/w3c/wot-discovery/issues/16#issuecomment-792663791
- [x] Change spec document to match header-based approach

***
<a href="https://pr-preview.s3.amazonaws.com/farshidtz/wot-discovery/pull/130.html" title="Last updated on Mar 23, 2021, 10:17 AM UTC (d78b6ca)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-discovery/130/e6f18dd...farshidtz:d78b6ca.html" title="Last updated on Mar 23, 2021, 10:17 AM UTC (d78b6ca)">Diff</a>

### DIFF
--- a/directory.td.json
+++ b/directory.td.json
@@ -210,8 +210,8 @@
                 "sort_order": {
                     "description": "Sorting order",
                     "type": "string",
-                    "enum": ["ASC", "DESC"],
-                    "default": "ASC"
+                    "enum": ["asc", "desc"],
+                    "default": "asc"
                 }
             },
             "forms": [
--- a/directory.td.json
+++ b/directory.td.json
@@ -202,12 +202,12 @@
                     "type": "boolean",
                     "default": false
                 },
-                "sortBy": {
+                "sort_by": {
                     "description": "The TD attribute used as comparator for sorting",
-                    "type": "boolean",
-                    "default": false
+                    "type": "string",
+                    "default": "id"
                 },
-                "sortOrder": {
+                "sort_order": {
                     "description": "Sorting order",
                     "type": "string",
                     "enum": ["ASC", "DESC"],
@@ -216,12 +216,20 @@
             },
             "forms": [
                 {
-                    "href": "/td{?offset,limit,count,sortBy,sortOrder}",
+                    "href": "/td{?offset,limit,count,sort_by,sort_order}",
                     "htv:methodName": "GET",
                     "response": {
                         "description": "Success response",
                         "htv:statusCodeValue": 200,
-                        "contentType": "application/ld+json"
+                        "contentType": "application/ld+json",
+                        "htv:headers": [
+                            {
+                                "htv:fieldName": "Link"
+                            },
+                            {
+                                "htv:fieldName": "Content-Range"
+                            }
+                        ],
                     },
                     "scopes": "readAll"
                 }
--- a/directory.td.json
+++ b/directory.td.json
@@ -17,6 +17,7 @@
             "scopes": [
                 "write",
                 "read",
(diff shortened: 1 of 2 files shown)


directory.td.json
```diff
@@ -181,6 +184,64 @@
...
+            "uriVariables": {
+                "offset": {
+                    "description": "The number of TDs to skip before the collection",
+                    "type": "number",
+                    "default": 0
+                },
+                "limit": {
+                    "description": "The number of TDs in the collection",
+                    "type": "number",
+                    "default": 1
```
farshidtz, line 198: @sebastiankb @mmccool Is there a way to set default values in TD? 
The [URI Template](https://tools.ietf.org/html/rfc6570) does not support it after [draft v5](https://tools.ietf.org/html/draft-gregorio-uritemplate-05#section-2.5) and I couldn't find anything in the TD's DataSchema either.

AndreaCimminoArriaga (approved):
I'm approving the review, as only remark I would add the optional functionality to return in the header the pair "Link" and an ontology class URI (as specified in the LDPP). In this way when the TDs are returned by using the headers one could know the types of these TDs.

farshidtz:
We could try to fully comply with [RFC7233 Range Requests](https://tools.ietf.org/html/rfc7233). This would let us query range of bytes (or range of TDs) by setting equivalent request headers.

To have dedicated resource URLs for the RESTful API, we could allow passing range request values in query. E.g.:
```
/td?range-start=0&range-end=100&range-unit=TDs gives the first 100 TDs
/td?range-start=0&range-end=100&range-unit=bytes gives the first 100 bytes
```
The above is inline with [LDP Paging](https://www.w3.org/TR/ldp-paging/). Moreover, we could borrow LDPP's Links spec to have next link and etag to identify the state.

AndreaCimminoArriaga:
@farshidtz it makes sense to have both ranges, maybe we could even define others. I like the approach, it is sensible.

index.html
```diff
@@ -1066,25 +1141,86 @@ <h4>Listing</h4>
...
+                                constructing the URL or by using the `next` Link from the previous response:
+                                <pre>
+                                    GET /td?offset=10&limit=10 HTTP/1.1   
+                                    Host: tdd.example.com
+                                </pre>
+
+                                Response:
+                                <pre>
+                                    HTTP/1.1 200 OK
+                                    Content-Range: TDs 10-12/*
```
danielpeintner, line 1192: I believe the range should be "10-11"
Otherwise it would be 3 more TDs and the count 12 does not match (0-11)

farshidtz:
I changed this to draft. As mentioned in my last [comment](https://github.com/w3c/wot-discovery/pull/130#issuecomment-801005408), we should try to fully comply with one standard and define query arguments to make pages addressable by URL. But this makes the specification even longer and this is overkill when there are much simpler approaches; see #145.

directory.td.json
mmccool: Best to create an issue about this in the TD repo, otherwise it may not get addressed.

mmccool:
Closing without merging, but as discussed Farshid will create a new PR with equivalent (but improved) content.

directory.td.json
farshidtz: Thanks. I'll reply on the new issue: https://github.com/w3c/wot-thing-description/issues/1093

PR #136, opened 2021-03-13
Validation error example

farshidtz:
This PR add a validation error example.

~~Moreover, it makes the use of [RFC7807 Problem Details](https://tools.ietf.org/html/rfc7807) recommended, instead of mandatory.~~

***
<a href="https://pr-preview.s3.amazonaws.com/farshidtz/wot-discovery/pull/136.html" title="Last updated on Mar 23, 2021, 3:33 PM UTC (e8022ef)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-discovery/136/329756e...farshidtz:e8022ef.html" title="Last updated on Mar 23, 2021, 3:33 PM UTC (e8022ef)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -654,7 +654,7 @@ img.wot-diagram {
                     The HTTP API responses must use appropriate status codes described in 
                     this section for success and error responses.
                     <span class="rfc2119-assertion" id="tdd-http-errors"> 
-                        The HTTP API SHOULD use the Problem Details [[RFC7807]] format to carry
+                        The HTTP API MUST use the Problem Details [[RFC7807]] format to carry
                         error details in HTTP client error (4xx) and server error (5xx) responses.
                     </span>
                     This enables both machines and humans to know the high-level error class
@@ -1133,11 +1133,11 @@ img.wot-diagram {
                             </span>
     
                             <span class="rfc2119-assertion" id="td-validation-response">
-                                The validation error SHOULD be described as Problem Details [[RFC7807]]
-                                and extended with a `validationErrors` array of objects
-                                with `field` and `description` fields to represent the error in a
-                                machine-readable way. 
+                                The validation error MUST be described as Problem Details [[RFC7807]]
+                                with an extension field called `validationErrors`, set to an array of objects
+                                with `field` and `description` fields.
                             </span>
+                            This is necessary to represent the error in a machine-readable way. 
                         </p>
                         
                         [[[#example-validation-error]]] is an example error response with two validation errors.
--- a/index.html
+++ b/index.html
@@ -1132,12 +1132,10 @@ img.wot-diagram {
                                 with necessary details to identify and resolve the errors.
                             </span>
     
-                            Validation errors are described as Problem Details [[RFC7807]], similar to all other
-                            response errors.
                             <span class="rfc2119-assertion" id="td-validation-response">
                                 The validation error SHOULD be described as Problem Details [[RFC7807]]
-                                and contain an `invalid-params` field set to an array of objects
-                                with `name` and `reason` fields to represent the error in a
+                                and extended with a `validationErrors` array of objects
+                                with `field` and `description` fields to represent the error in a
                                 machine-readable way. 
                             </span>
                         </p>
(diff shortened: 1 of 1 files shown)


mmccool:
Thanks for the example.  I would consider this editorial except for the change to the assertion regarding RFC7807 from MUST to SHOULD.  To maximize interoperability I would have thought we should specify the data schema for returned error information, and if it is mandatory to format error data a certain way then this should be a MUST.  What is the reasoning behind converting this to a SHOULD?

index.html
```diff
@@ -1115,25 +1115,60 @@ <h4>Listing</h4>
...
-                        </span>
-                        <span class="rfc2119-assertion" id="td-validation-result">
-                            If the server fails to validate the TD object, it MUST inform the client
-                            with necessary details to identify and resolve the errors.
-                        </span>
-
-                        <p class="ednote" title="Validation Example">
-                            It would be nice to include an example validation result in
-                            the uniform error format (i.e. RFC7807 Problem Details).
+                        <p>
```
mmccool, line 1118: This is ok for now but I'm hoping to clean this up in the near future and add "Syntactic" and "Semantic" validation definitions in the TD specification.   After that is done we can replace this with a reference to those sections.

index.html
```diff
@@ -654,7 +654,7 @@ <h3>Directory Service API</h3>
                     The HTTP API responses must use appropriate status codes described in 
                     this section for success and error responses.
                     <span class="rfc2119-assertion" id="tdd-http-errors"> 
-                        The HTTP API MUST use the Problem Details [[RFC7807]] format to carry
+                        The HTTP API SHOULD use the Problem Details [[RFC7807]] format to carry
```
mmccool, line 657: I'm not convinced this is a good idea.  Don't we want to be clear what the data schema of all data returned by the directory is?  What is the justification for allowing implementations to use alternative error formats?  Note: I think this should be a MUST, but I think it would be ok (regardless of what RFC7807 says, I need to check) to allow *extensions* of the problem details format, e.g. implementation-dependent extra fields.

index.html
```diff
@@ -1115,25 +1115,60 @@ <h4>Listing</h4>
...
                         </p>
+                        <p>
+                            <span class="rfc2119-assertion" id="td-validation-result">
+                                If the server fails to validate the TD object, it MUST inform the client
+                                with necessary details to identify and resolve the errors.
+                            </span>
+    
+                            <span class="rfc2119-assertion" id="td-validation-response">
+                                The validation error SHOULD be described as Problem Details [[RFC7807]]
+                                and extended with a `validationErrors` array of objects
```
mmccool, line 1137: ... with an extension...; is there a data schema for this in the TD?

index.html
```diff
@@ -1115,25 +1115,60 @@ <h4>Listing</h4>
...
+                            </span>
                         </p>
+                        <p>
+                            <span class="rfc2119-assertion" id="td-validation-result">
+                                If the server fails to validate the TD object, it MUST inform the client
+                                with necessary details to identify and resolve the errors.
+                            </span>
+    
+                            <span class="rfc2119-assertion" id="td-validation-response">
+                                The validation error SHOULD be described as Problem Details [[RFC7807]]
```
mmccool, line 1136: As above, please justify why this is a SHOULD rather than a MUST

farshidtz:
> Thanks for the example. I would consider this editorial except for the change to the assertion regarding RFC7807 from MUST to SHOULD. To maximize interoperability I would have thought we should specify the data schema for returned error information, and if it is mandatory to format error data a certain way then this should be a MUST. What is the reasoning behind converting this to a SHOULD?

I think the response codes are often sufficient to identify errors. People had disagreements with mandating of the schema for error responses from the beginning and I noticed one implementation ignoring this assertion. In any case, I can make a commit to revert that and we can discuss it in a separate issue.

index.html
farshidtz: Not yet. It depends in being able to define an schema for additionalResponses: https://github.com/w3c/wot-thing-description/issues/1053

index.html
farshidtz: Replied below.

index.html
mmccool: We basically agreed that we would support this in TDs although I need to finalize the PR for it so it can be merged.  Rather than putting the data schema in the form directly they will be in a named map like securityDefinitions, though.

mmccool:
Let's leave the assertion as a MUST for now so we can consider this PR editorial.  Can you create an issue about the change from MUST to SHOULD?  I personally think it's ok to omit return data in an error but if data IS returned, it MUST follow an expected format and not something random.  So maybe the objection people had was being forced to always return something.  I have to look in detail to what the RFC says, and also figure out how we can indicate alternative responses one of which returns data and one that does not in the data schemas... maybe "oneOf" and the "null" data schema?

Issue #149, opened 2021-04-09
Anonymous TDs in a directory


farshidtz:
Anonymous TD are those which don't have an `id` field. 

Should the directory accept them? If yes, how should they be maintained and exposed anonymously?

farshidtz:
For the HTTP API:
- Submission: With a POST request. System-generated pseudorandom (e.g. UUIDv4 URN) ID given to producer in response. 
  - Should the system-generated ID be added to the resource? If yes, will the TD remain anonymous?
  - How to avoid duplicates? See #48
- Retrieval: With the system-generated ID.
- Listing (retrieval of all TDs): Returning anonymous TDs along with others. How to sort by ID (default) when anonymous entries have no IDs?
- Search: Performing search on the full list. Same issue as for listing.
- Notifications: Create/Update/Delete events need to provide the TD's identifier. Exposing system-generated IDs there suffers from the same issue associated with adding the ID inside the TD.

egekorkan:
I think that the comments of the privacy group (PING) regarding having mandatory id in the previous versions of the TD would be applied here as well if an id is inserted by the system. Once a TD is in the  directory and has an id, it will be probably always consumed with the version from the directory no?

benfrancis:
I nearly filed an issue about this a few weeks ago because I noticed the Directory Service API needs an ID to fetch a Thing Description, but `id` is not a mandatory member of a Thing Description. That could mean that when you fetch a list of Thing Descriptions from a directory, it isn't possible to determine the URL of an individual Thing Description in the directory if it doesn't provide an `id` member in its listing.

Then I noticed that the specification said the directory would provide a system-generated ID for "anonymous" Thing Descriptions, so that shouldn't be a problem (I assumed those IDs would also be provided as an `id` member in the directory listing).

Note that this does mean the Thing Description exposed by the directory wouldn't be exactly the same as the Thing Description provided by the client (as it would have an additional `id` member), which I personally think is fine, but some people have expressed a desire to add resource level signing to canonical Thing Descriptions which would be invalidated by such a modification.

[...]

PR #152, opened 2021-04-13
Add Registration ID to Enriched TD

farshidtz:
Related to #149.

I added the URL inside the registration object for now to be consistent with the other registration attributes. We could move them all to the top level and pick more suitable names.

```json
{
    "@context": [
        "http://www.w3.org/ns/td",
    ],
    "id": "urn:example:simple-td",
    "security": "basic_sc",
    "securityDefinitions": {
        "basic_sc": {
            "scheme": "basic"
        }
    },
    "tdd:registration": {
        "tdd:created": "2021-01-19T17:02:00Z",
        "tdd:href": "https://tdd.example.com/td/urn:example:simple-td",
        "tdd:modified": "2021-01-19T17:02:00Z"
    },
    "title": "Simple TD"
}
```

Also, I made it explicit that retrieved TDs will be enriched. This is necessary to identify Anonymous TDs in listing and search APIs.

***
<a href="https://pr-preview.s3.amazonaws.com/farshidtz/wot-discovery/pull/152.html" title="Last updated on Apr 19, 2021, 11:08 AM UTC (6030b4c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-discovery/152/fab5087...farshidtz:6030b4c.html" title="Last updated on Apr 19, 2021, 11:08 AM UTC (6030b4c)">Diff</a>


farshidtz:
Looking at the sorting and notification requirements, I think adding an `id` instead of `href` will be more appropriate. This will be same as what we originally agreed on but removed because `@id` of `tdd:registration` object was semantically the id of registration object and nothing related to the TD. One solution to alias `tdd:id` to something like thingRegistrationID instead of `@id`. 

Reasons for `id` instead of `href`:
- Sorting by a shorter string will be computationally less complex than sorting by `href`.
- Construction of an absolute URL requires knowledge of the fully qualified domain name at the time of deployment and will prevent serving via both local and public endpoints.

```json
{
    "@context": [
        "http://www.w3.org/ns/td",
    ],
    "id": "urn:example:simple-td", // anonymous TDs will not have this
    "security": "basic_sc",
    "securityDefinitions": {
        "basic_sc": {
            "scheme": "basic"
        }
    },
    "tdd:registration": {
        "tdd:created": "2021-01-19T17:02:00Z",
        "tdd:href": "https://tdd.example.com/td/urn:example:simple-td",
        "tdd:id": "urn:example:simple-td", // id instead of href?
        "tdd:modified": "2021-01-19T17:02:00Z"
    },
    "title": "Simple TD"
}
```

farshidtz:
I just had a call with @AndreaCimminoArriaga to discuss the proposal. We came up with:

Context file based on what we already have in the ontology:
```json
{
  "@context": {
    "discovery": "https://www.w3.org/2021/wot/discovery#",
    "DirectoryDescription": {
      "@id": "discovery:DirectoryDescription"
    },
    "LinkDescription": {
      "@id": "discovery:LinkDescription"
    },
    "RegistrationInformation": {
      "@id": "discovery:RegistrationInformation"
    },
    "id": "@id",
    "type": "@type",
    "registration": "discovery:hasRegistrationInformation",
    "created": "dc:created",
    "modified": "dc:modified"
  }
}
```

Known TD:
```json
{
    "@context": [
        "http://www.w3.org/ns/td",
    ],
    "id": "urn:example:known-td",
    "security": "basic_sc",
    "securityDefinitions": {
        "basic_sc": {
            "scheme": "basic"
        }
    },
    "registration": {
        "created": "2021-01-19T17:02:00Z",
        "id": "https://tdd.example.com/td/urn:example:known-td", 
        "modified": "2021-01-19T17:02:00Z"
    },
    "title": "Known TD"
}
```

Anonymous TD:
```json
{
    "@context": [
        "http://www.w3.org/ns/td",
    ],
    "security": "basic_sc",
    "securityDefinitions": {
        "basic_sc": {
            "scheme": "basic"
        }
    },
    "registration": {
        "created": "2021-01-19T17:02:00Z",
        "id": "https://tdd.example.com/td/urn:uuid:ab90172b-c3b7-4f71-9ea0-2dc7cf9b796f", 
        "modified": "2021-01-19T17:02:00Z"
    },
    "title": "Anonymous TD"
}
```

We can move all the registration attributes to TD itself and give them different names. That way, we would become inline with @vcharpenay's suggestion at https://github.com/w3c/wot-discovery/issues/149#issuecomment-817746786. Signature verification can still work either by specifying what is signed (https://github.com/w3c/wot-thing-description/issues/940#issuecomment-816570048) or by removing (unenriching) the TD before validation. Moreover, this makes the notification object simpler (more on that in a future PR).

farshidtz:
There is a recommended way to add local identifiers to blank nodes (in this case the Anonymous TD):
https://json-ld.org/spec/latest/json-ld/#identifying-blank-nodes

An identifier that is not an absolute URL also brings the following benefits:
- Sorting by a shorter string will be computationally less complex than sorting by url. We can still recommend servers to sort by the variable part of it, but we need to describe that variable and maybe have it in the ontology too.
- Construction of an absolute URL requires knowledge of the fully qualified domain name (FQDN) at the time of deployment and will prevent serving via both local or more than one public (proxied) endpoints.

index.html
```diff
@@ -773,8 +824,10 @@ <h4>Creation</h4>
                                     and a Location header containing a system-generated identifier for the TD.
                                 </span>
                                 <span class="rfc2119-assertion" id="tdd-reg-create-anonymous-td-generated-id"></span>
-                                    The identifier SHOULD be a Version 4 UUID URN [[RFC4122]].
+                                    The identifier SHOULD be a UUID Version 4 [[RFC4122]].
                                 </span>
+                                That is a random or pseudo-random number which does not carry unintended information
+                                about the host or the resource.
```
farshidtz, line 830: @mmccool should we add some recommendation about generation of secure identifiers in the privacy and security section? In particular, UUID versions other than v4 may need to be avoided.

mmccool: please create an issue about this in wot-security and we will discuss

PR #124, opened 2021-02-28
Registration spec: listing

farshidtz:
The discussion is still ongoing in #16. This PR adds the "draft" of what we have agreed on in terms of paginated listing.

***
<a href="https://pr-preview.s3.amazonaws.com/farshidtz/wot-discovery/pull/124.html" title="Last updated on Mar 1, 2021, 1:05 PM UTC (e6f18dd)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-discovery/124/d2edad7...farshidtz:e6f18dd.html" title="Last updated on Mar 1, 2021, 1:05 PM UTC (e6f18dd)">Diff</a>

Issue #117, opened 2021-02-11
Incremental TDs provided by constrained devices


mmccool:
Constrained devices should be able to incrementally provide and register TDs

mmccool:
on a 64K device we may only have a few K free to send a TD, so may have to pull it out of flash and fill in blanks (and compute the signature!)

mmccool:
right now we are going to define profiles and reference platforms so the entire TD can fit in memory (eg 256K) for processing prior to being provided for self-description and directory registration but we should design our network interfaces so smaller devices are *possible*...

farshidtz:
An alternative is to have those constrained devices only self-describe or register a minimal TD with a link (see https://github.com/w3c/wot-discovery/issues/34#issuecomment-780393770) pointing to the complete TD hosted in a directory (registered manually or by a more powerful device) or elsewhere.

farshidtz:
> Constrained devices should be able to incrementally provide and register TDs

I think this should be handled at the protocol level. HTTP requests and responses can be transmitted in chunks (see https://en.wikipedia.org/wiki/Chunked_transfer_encoding). But HTTP/2 doesn't support that and instead offers framing.

benfrancis:
> I think this should be handled at the protocol level. HTTP requests and responses can be transmitted in chunks (see https://en.wikipedia.org/wiki/Chunked_transfer_encoding).

See also: https://developer.mozilla.org/en-US/docs/Web/HTTP/Range_requests

farshidtz:
> > I think this should be handled at the protocol level. HTTP requests and responses can be transmitted in chunks (see https://en.wikipedia.org/wiki/Chunked_transfer_encoding).
> 
> See also: https://developer.mozilla.org/en-US/docs/Web/HTTP/Range_requests

@benfrancis I understand that Range requests should be initiated by the client. Do you know of any specification for a server to refuse full (normal) requests by clients and only respond to partial requests? Will returning [416 Range Not Satisfiable](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/416) make sense in such situation? It is meant for an unsatisfiable range. Or [413 Payload Too Large](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/413), considering that payload is too large for the request? I am more tempted toward 416.

I'm asking because in the case of limited devices, the device has no way of returning the whole resource at once if the clients asks for it.

benfrancis:
[...]

PR #153, opened 2021-04-13
Listing with pagination

farshidtz:
This will add pagination to the spec. It includes the minimum requirements taken from #130.

This is compatible with the existing listing mechanism.

Note: This depends on #152 and should not be merged beforehand.

Todos:
- [x] Pagination rules
- [x] Example
- [x] Spec in TD

Other related issues/PRs: #16, #145, #149

***
<a href="https://pr-preview.s3.amazonaws.com/farshidtz/wot-discovery/pull/153.html" title="Last updated on Apr 19, 2021, 11:08 AM UTC (768d1f6)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-discovery/153/fab5087...farshidtz:768d1f6.html" title="Last updated on Apr 19, 2021, 11:08 AM UTC (768d1f6)">Diff</a>

### DIFF
--- a/directory.td.json
+++ b/directory.td.json
@@ -230,6 +230,28 @@
         },
         "retrieveTDs": {
             "description": "Retrieve all Thing Descriptions",
+            "uriVariables": {
+                "offset": {
+                    "title": "Number of TDs to skip before the page",
+                    "type": "number",
+                    "default": 0
+                },
+                "limit": {
+                    "title": "Number of TDs in a page",
+                    "type": "number"
+                },
+                "sort_by": {
+                    "title": "Comparator TD attribute for collection sorting",
+                    "type": "string",
+                    "default": "id"
+                },
+                "sort_order": {
+                    "title": "Sorting order",
+                    "type": "string",
+                    "enum": ["asc", "desc"],
+                    "default": "asc"
+                }
+            },
             "forms": [
                 {
                     "href": "/td",
@@ -237,8 +259,30 @@
                     "response": {
                         "description": "Success response",
                         "htv:statusCodeValue": 200,
-                        "contentType": "application/ld+json",
+                        "contentType": "application/ld+json"
                     },
+                    "scopes": "readAll"
+                },
+                {
+                    "href": "/td{?offset,limit,sort_by,sort_order}",
+                    "htv:methodName": "GET",
+                    "response": {
+                        "description": "Success response",
+                        "htv:statusCodeValue": 200,
+                        "contentType": "application/ld+json",
+                        "htv:headers": [
+                            {
+                                "htv:fieldName": "Link"
+                            }
+                        ]
+                    },
+                    "additionalResponses": [
+                        {
+                            "description": "Invalid query arguments",
+                            "contentType": "application/problem+json",
+                            "htv:statusCodeValue": 400
+                        }
+                    ],
                     "scopes": "readAll"
                 }
             ]
--- a/index.html
+++ b/index.html
@@ -964,8 +964,8 @@ img.wot-diagram {
                         <h4>Listing</h4>
                         <p>
                             The listing endpoint provides a way to query the collection of TD objects
-                            from the directory. The Search API may be used to retrieve <a>Partial TD</a>s
-                            or TD fragments; see [[[#exploration-directory-api-search]]].
+                            from the directory. The Search API may be used to retrieve TD fragments; 
+                            see [[[#exploration-directory-api-search]]].
                         </p>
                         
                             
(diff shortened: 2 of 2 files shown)


sebastiankb:
Is there a good summary somewhere of why the earlier version of pagination was removed?

farshidtz:
> Is there a good summary somewhere of why the earlier version of pagination was removed?

I assume you are referring to #124 which was replaced by #145. There were a few issues discussed in the discovery calls and in linked issues. 

I will summarize what I remember:
- The envelope wasn't valid from an ontology perspective (see https://github.com/w3c/wot-discovery/issues/54#issuecomment-792660587 and  https://github.com/w3c/wot-discovery/issues/16#issuecomment-792663791) and we received recommendation from TD ontology authors to follow [Linked Data Platform Paging 1.0](https://www.w3.org/TR/ldp-paging/). This current PR is based on it; I have to mention it somewhere.
- The pagination alone isn't sufficient to reduce load on servers/clients because the resources are of variable size.
- The envelope object around the array was inconsistent with the search APIs response format. All of jsonpath, xpath, and sparql search APIs return arrays. By returning a similar format, we simplify the specification and allow clients to receive the whole collection or a filtered subset in the same payload format. Not fully relevant, but we are now looking into merging syntactic search and listing APIs (details in #133).

wiresio:
Thanks for summarizing @farshidtz !
After having read (again) through the issues mentioned and your comment I'm more confused than before:
- XPath and SPARQL queries usually do not return (JSON) arrays, see e.g.: https://www.w3.org/TR/2013/REC-sparql11-results-json-20130321/#json-result-object
- So anyway, the TDD frontend has to convert the response to another format, in our case JSON (array) for the search APIs.
- Why do we then go for JSON-LD at the listing API? Could also be plain JSON allowing more degrees of freedom.
- Phrases like "HTTP/1.1 servers SHOULD perform chunked Transfer-Encoding" are useful hints for developers but should not prescribe the individual implementation. Reducing load for servers/clients is IMHO highly dependent on the environment you're in. If a clever format, like JSON-LD / array of TDs, helps here, than we should emphasize it instead of insisting on protocol specifics

Sorry for potentially discussing the same stuff again and again, but maybe I've just missed some basic assumptions like "responses should be JSON-LD in case ... because they help / allow for ...".

farshidtz:
> * XPath and SPARQL queries usually do not return (JSON) arrays, see e.g.: https://www.w3.org/TR/2013/REC-sparql11-results-json-20130321/#json-result-object

Sorry that was a wrong statement (at least the sparql part). But for JSONPath and XPath, the result of executing the expressions to do filtering on the collection will be an array. There could be other formats on function calls for e.g. to count elements. We need to go back to the Search API spec and check if we have claimed something wrong. But that is not relevant to this PR.

> * Why do we then go for JSON-LD at the listing API? Could also be plain JSON allowing more degrees of freedom.

It is (to my limited understanding of linked data), a bad practice to embed JSON-LD (i.e. an RDF document) in plain JSON. @AndreaCimminoArriaga has discussed this in the calls few times.
Note: This PR is not suggesting a new payload.

> * Phrases like "HTTP/1.1 servers SHOULD perform chunked Transfer-Encoding" are useful hints for developers but should not prescribe the individual implementation. Reducing load for servers/clients is IMHO highly dependent on the environment you're in. If a clever format, like JSON-LD / array of TDs, helps here, than we should emphasize it instead of insisting on protocol specifics

That is not part of this PR. Anyway, those are recommendations based on HTTP specification to allow incremental transfer of possibly large TDs in IoT and other environments. We had a lengthy discussion in #145. The same recommendation should be added for retrieval or submission of one TD (#117). The servers are NOT mandated to follow it. 

The data model, serialization format, and transfer encoding are different topics. The protocol-specific statements for HTTP/1.1 and HTTP/2 are to do efficient transfer on the named protocols.

> Sorry for potentially discussing the same stuff again and again, but maybe I've just missed some basic assumptions like "responses should be JSON-LD in case ... because they help / allow for ...".

I think the reason is same as why the TD should be JSON-LD.

wiresio:
> It is (to my limited understanding of linked data), a bad practice to embed JSON-LD (i.e. an RDF document) in plain JSON. @AndreaCimminoArriaga has discussed this in the calls few times.

Thanks & apologies again for (too much) insisting. That was my missing part. @AndreaCimminoArriaga - please feel free to comment or point me to the discussion you already had. Would be great to have streamlined search APIs and even better to have both search and listing APIs streamlined.

wiresio:
@egekorkan - I know that you are working on System Descriptions. Did you ever come across the issue of embedding TDs having their own context inside a JSON-LD frame having another context or just a plain JSON? If yes, how did you solve that? Thanks!

egekorkan:
@wiresio so in SDs there are TD fragments that are of interest for a system and they identified in the SD ontology with the TD context: https://github.com/tum-esi/wot-system-description/blob/b45b06f97da8aca507a047865b1148d2fb3357a6/paper-mentions/3_Context/sdContext.json#L22
You can find the ontology here as well: https://github.com/tum-esi/wot-system-description/tree/master/paper-mentions/3_Context

However, I think that I do not answer your question exactly :/

wiresio:
@egekorkan - Thanks for pointing me to your work!  Looks like it is not directly applicable for TDD purposes ... was just a thought that came into my mind.

PR #159, opened 2021-04-22
Split registration event into multiple events

benfrancis:
This PR proposes splitting the `registration` event of the Directory Service API into three separate events: `thingCreated`, `thingUpdated` and `thingDeleted`. As discussed in #133.

The rationale for combining these three event types into a single event in the first place was to make it easier to subscribe to all events at once. But this feels like an anti-pattern which is motivated by needing to work around limitations of the WoT Thing Description specification. It also makes it impossible for a Consumer to subscribe to individual events if that's what they want to do.

A normative Thing Description or Thing Model specified by the WoT Working Group should demonstrate best practices, and taking the combined approach may imply that it's a good practice for Web Things to combine all their event types into a single event to make it easier to subscribe to all events at the same time.

I have instead proposed new `subscribeallevents` and `unsubscribeallevents` operations for the Thing Description specification so that it could be possible to either subscribe to all events, or to individual events, depending on what the Consumer prefers.  See #1082.

***
<a href="https://pr-preview.s3.amazonaws.com/benfrancis/wot-discovery/pull/159.html" title="Last updated on Apr 30, 2021, 4:24 PM UTC (e35690e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-discovery/159/581dacc...benfrancis:e35690e.html" title="Last updated on Apr 30, 2021, 4:24 PM UTC (e35690e)">Diff</a>

### DIFF
--- a/directory.td.json
+++ b/directory.td.json
@@ -389,43 +389,106 @@
         }
     },
     "events": {
-        "registration": {
-            "description": "TD registration events",
-            "uriVariables": {
-                "type": {
-                    "title": "Event type(s)",
-                    "description": "Multiple types passed as type={type} pairs for SSE",
-                    "type": "string",
-                    "enum": [
-                        "create",
-                        "update",
-                        "delete"
-                    ]
-                },
-                "full": {
-                    "title": "Include TD changes inside event data",
-                    "type": "boolean"
-                }
-            },
-            "forms": [
-                {
-                    "op": "subscribeevent",
-                    "href": "/events{?type,full}",
-                    "subprotocol": "sse",
-                    "contentType": "text/event-stream",
-                    "htv:headers": [
-                        {
-                            "description": "ID of the last event provided by reconnecting clients",
-                            "htv:fieldName": "Last-Event-ID"
-                        }
-                    ],
-                    "data": {
-                        "description": "Partial or complete TD",
-                        "type": "object"
-                    },
-                    "scopes": "notifications"
-                }
-            ]
+        "thingCreated": {
+          "@type": "ThingCreatedEvent",
+          "title": "Thing created",
+          "data": {
+              "type": "object",
+              "description": "The schema of created TD event data including the created TD",
+              "properties": {
+                  "td_id": {
+                      "type": "string",
+                      "format": "iri-reference",
+                      "description": "Identifier of TD in directory"
+                  },
+                  "td": {
+                      "type": "object",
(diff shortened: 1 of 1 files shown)


farshidtz:
> It also makes it impossible for a Consumer to subscribe to individual events if that's what they want to do.

They can subscribe to individual events. I guess you meant it is impossible to un-subscribe from individual events?

Though that is also possible. A client can unsubscribe separately if they have subscribed the same way.

farshidtz:
The event format and query arguments were changed recently. It would be great if you could propose your changes based on the latest spec.

mmccool:
Query approach is more extensible (can add new event types without changing the API) and maps better to MQTT (point from @farshidtz).

farshidtz:
Example new event type given by @mmccool: TD expiry.

benfrancis:
@mmccool Do you think it's good practice to combine all event types that a device emits into a single event affordance?

I would argue adding an event type changes the API regardless of the URL structure used to subscribe to events. Either way a Consumer still need modifying to support the new event type.

Regarding MQTT, see my comment at https://github.com/w3c/wot-discovery/pull/158#issuecomment-832105679

farshidtz:
> I would argue adding an event type changes the API regardless of the URL structure used to subscribe to events. Either way a Consumer still need modifying to support the new event type.

This is small technical issue and indeed a good point. We should write in the spec that servers can add other events types and clients should ignore anything that they don't recognize.

> Regarding MQTT, see my comment at [#158 (comment)](https://github.com/w3c/wot-discovery/pull/158#issuecomment-832105679)

The current PR is about notifications which is very well relevant to MQTT. I think most implementations will end up using MQTT instead of SSE but we didn't specify that because we wanted to limit the first version to HTTP only and simplify the security (as far as I recall).

To expand my point about mapping to MQTT. There is no negotiation capability to request full or filter in MQTT, but the event type query will map this way:
SSE (HTTP Path) `/events{?type}` -> MQTT Topic `events{/type}`

mmccool:
- actually opposite of consolidation...
- having separate interactions is more convenient in scripting API
- having single URL with query interface more convenient when working directly with URLs 
- no reason we can't have both! - separate interactions, using common path/query interface underneath

mmccool:
@farshidtz notes that this is based on an old version of the spec... so we can't necessarily merge at present

benfrancis:
Superseded by #184

PR #160, opened 2021-04-22
Rename actions to match properties and events

benfrancis:
This PR proposes renaming and combining actions and their API endpoints in the Directory Service API to match other proposed changes to the example Thing Directory Description (in #158 and #159):
- `createTD` -> `createThing`
- ~~`updateTD` & `updatePartialTD` -> `updateThing`~~
- `updateTD` -> `updateThing`
- `updatePartialTD` -> `partialUpdateThing`
- `deleteTD` -> `deleteThing`

See discussion in #133

***
<a href="https://pr-preview.s3.amazonaws.com/benfrancis/wot-discovery/pull/160.html" title="Last updated on May 18, 2021, 2:20 PM UTC (96cd40f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-discovery/160/581dacc...benfrancis:96cd40f.html" title="Last updated on May 18, 2021, 2:20 PM UTC (96cd40f)">Diff</a>

### DIFF
--- a/directory.td.json
+++ b/directory.td.json
@@ -53,10 +53,12 @@
     "security": "combo_sc",
     "base": "https://tdd.example.com",
     "actions": {
-        "createTD": {
+        "createThing": {
+            "@type": "CreateThingAction",
             "description": "Create a Thing Description",
             "uriVariables": {
                 "id": {
+                    "@type": "ThingID",
                     "title": "Thing Description ID",
                     "type": "string",
                     "format": "iri-reference"
@@ -64,7 +66,7 @@
             },
             "forms": [
                 {
-                    "href": "/td/{id}",
+                    "href": "/things/{id}",
                     "htv:methodName": "PUT",
                     "contentType": "application/td+json",
                     "response": {
@@ -81,14 +83,14 @@
                     "scopes": "write"
                 },
                 {
-                    "href": "/td",
+                    "href": "/things",
                     "htv:methodName": "POST",
                     "contentType": "application/td+json",
                     "response": {
                         "description": "Success response",
                         "htv:headers": [
                             {
-                                "description": "System-generated UUID (version 4) URN",
+                                "description": "System-generated URI",
                                 "htv:fieldName": "Location",
                                 "htv:fieldValue": ""
                             }
@@ -106,10 +108,11 @@
                 }
             ]
         },
-        "updateTD": {
+        "updateThing": {
             "description": "Update a Thing Description",
             "uriVariables": {
                 "id": {
+                    "@type": "ThingID",
                     "title": "Thing Description ID",
                     "type": "string",
                     "format": "iri-reference"
@@ -117,7 +120,7 @@
             },
             "forms": [
                 {
-                    "href": "/td/{id}",
+                    "href": "/things/{id}",
                     "htv:methodName": "PUT",
                     "contentType": "application/td+json",
                     "response": {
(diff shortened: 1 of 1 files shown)


directory.td.json
```diff
@@ -106,18 +108,19 @@
                 }
             ]
         },
-        "updateTD": {
+        "updateThing": {
```
farshidtz, line 111: Missing @type semantic annotation.

directory.td.json
```diff
@@ -169,18 +160,19 @@
                 }
             ]
         },
-        "deleteTD": {
+        "deleteThing": {
```
farshidtz, line 163: Missing @type.

directory.td.json
```diff
@@ -81,14 +83,14 @@
...
                     "htv:methodName": "POST",
                     "contentType": "application/td+json",
                     "response": {
                         "description": "Success response",
                         "htv:headers": [
                             {
-                                "description": "System-generated UUID (version 4) URN",
+                                "description": "System-generated URI",
                                 "htv:fieldName": "Location",
                                 "htv:fieldValue": ""
```
farshidtz, line 95: This can be removed.

mmccool:
Not merging for now, some concern about merging PATCH with PUT.   It may be more RESTful, but is also more inconvenient since we don't have a good/convenient way to select forms in the scripting API.  Two paths forward:
1. Use actions (e.g. separate interactions) for updates as in the existing API
2. Fix the scripting API to allow easy selection of forms by methods (but... protocol specific).   Using different interactions (even if they map to the same URL in a given protocol) we can decouple the script from the protocol.

In general... making things more "RESTful" also seem to be making them more protocol specific and may be a step backwards.   Just having one name (interaction) per distinct API operations seems... simpler.

mmccool:
Also, to be clear: I think the TD for the directory should be a TM and normative...

benfrancis:
I understand what you're getting at. But the `PUT` and `PATCH` in this API both correspond to a `writeproperty` operation. If they are considered different interactions, should there also be a separate operation type for partial writes (e.g. `partialwriteproperty`) in order to distinguish between those different interaction affordances?

farshidtz:
> @mmccool I'm a bit confused about why it's OK for `createThing` to have both a `PUT` and a `POST` option, but `updateThing` can not have both a `PUT` and a `PATCH` option?

As @relu91 explained above, `PUT` and `PATCH` accept different content types and do very different operations.

For `createThing`, both `PUT` and `POST` create a TD and the difference is whether or not the ID is user-defined. A mistake by the client will most probably result in an error. If necessary, we could move `POST` to a new interaction called `createAnonymousTD`.

mmccool:
- PUT/POST: scripting API implementation (e.g. node-wot) needs to handle as part of registration; hidden from users.
- PUT/PATCH: we don't have a partialwrite op, and so we can't distinguish these two at the scripting API level just using interaction/op combination.  Technically you could use formIndex to pick one but it's ugly.   Two possible solutions:
    1. Use different interaction name for PUT and PATCH 
    2. Add an appropriate op type to the TD spec
    3. Add a hook to the scripting API to fix this

My opinion: 1 is the easy thing to do, 2 is the right thing to do, 3 is a hacky workaround.  We *could* do both 1 and 2 (e.g. doing 1 does not preclude doing 2 later).

Actions:
1. Create issue in TD spec about adding a partialwrite op to support PATCH (esp when in same interaction with a PUT).  Will also want to raise issue in Scripting API.
2. Split this PR into two: (1) does renamings we agree on; (2)(which we can defer until we decide we want it) that merges PUT and PATCH interactions.  @farshidtz has agreed to do (1).   If this is done carefully, then *this* PR will only have the changes for (2) left.
3. @farshidtz will also do a PR to separate POST and PUT for registration, to have a "CreateAnonymousThing" action; then it would be an error to try to use "CreateThing" for a TD with no ID.

benfrancis:
I've added a commit to re-combine update and partial update.

benfrancis:
Superseded by #184

Issue #18, opened 2020-04-20
Consider TTL Use cases


mmccool:
When is TTL mandatory?   When is it not?
Note: In LinkSmart, TTL is optional.
Ex: Registration is created by a human using a UI, expected that human will remove it when it is no longer valid.   But when added by a device, TTL should be mandatory.

farshidtz:
Elaborating the use-cases for optional TTL in Thing Description (TD) of LinkSmart's Thing Directory: 

In that context, the TTL is the lifespan of the TD registration after which the directory removes the entry in the absence of any keepalive request. In other words, the directory intends to keep TDs as long as the actual thing exists and removes TDs that have `time > modified+TTL`. The consumers may rely on TTL to know when (i.e. `modified+TTL`) this registration becomes obsolete, after which they have to discover the TD again to get the up to date metadata.

The keepalive request is currently sent by the registrator via PUT (full TD) or PATCH (partial TD / no body) request over the RESTful API. The PATCH support is WIP.

Use cases:

- Sensors that are incapable of registering themselves
E.g. a zwave smart plug with no IP networking capabilities
- WoT non-compliant things
E.g. a proprietary, closed z-wave gateway device

There are two ways to register such devices: 

1. Using a third-party script hosted elsewhere (e.g. gateway). In this case, the TTL may be set and TD keepalive routine may be handled by the script. But lifecycle should corresponds to the actual thing rather than the script.

2. Using the UI of the Thing Directory by a human operator. The operator is expected to remove the registration when the thing is no longer available. This may also work with a long TTL, enforcing the operator to visit the UI and press some kind of validation button. But allowing long TTL defeats the purpose of TTL and may be misinterpreted by developers and misused on machines.

mmccool:
Additional use case:
- Device is deinstalled by user or fails.  In this case TTL ensures the directory is (eventually) cleaned up.

Additional questions:
- Can a consumer find out the TTL of a TD, and when it was last updated?
- During discovery I might want to filter out TDs that have not been updated in a while.
- If TTL is mandatory, what should the upper bound be?  Should be finite and "reasonable" for a device that is sleeping
- Does "keepalive" just require a "touch" message or does the TD need to be fully reloaded?  (see PATCH above)
- Is "keepalive" possible to include in other messages, eg as a header option?   Note that not all/most communications with a device will involve the directory.

mmccool:
Regarding upper bound, we can look at related protocols like DNS.
[...]

PR #162, opened 2021-04-23
Expiry date for registered TDs

farshidtz:
Related to #18, this PR adds description of a new `expires` field inside Registration Info.

- The expiry has few uses and should perhaps be a TD attribute; see https://github.com/w3c/wot-thing-description/issues/916
- The removal of expired TD from directories is recommended.
- Producers and consumers are encouraged to use it to indicate validity of dynamic TDs. This is also useful for self-describing things.
- There is NO upper limit for the expiry date, but a short interval is suggested for devices that may not be able to deregister.

***
<a href="https://pr-preview.s3.amazonaws.com/farshidtz/wot-discovery/pull/162.html" title="Last updated on May 10, 2021, 10:18 AM UTC (a28a266)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-discovery/162/581dacc...farshidtz:a28a266.html" title="Last updated on May 10, 2021, 10:18 AM UTC (a28a266)">Diff</a>


context/discovery-ontology-chowlk.xml
```diff
@@ -1,58 +1,2 @@
...
...wvSpl25AdQ54ZsmvIniHPDXlhyL4uSwn6pMuOIbuGPDdkv2Nf+ocM/5DhHzL8Q4Z/yPAPGf4hwz9k+NfKyPAPGf4hwz9k+Ida/+B7kdifzV099r1+7CPL7ce+c4HQn/VCv8uo/5CIFoJjQVn2V7ZhPC0vR0yMmLguJhaWvh641gAmhtaDS2DCHUg8jIAnWbiUyRpIQYKLggZ6jJMDFd/LFXKhpB+dO3cHhZRSeK6FDPr5vSv8KC24tdhWK6W2Xp2GyrYwD5Qk7wF42bZJBMv2Q0gXlSOMi5hFLMPJfatV0yfLaZNXsB0PiLZgCswjIjQiOWOKO1PoDsxgreMkAbJ50jsxNK2qhYpQWladGRFkG5FR+aNqddNQw5DnzE8bqkahZ6iMssbtswLPO5rxFjnOtAj0/t0xoSVqNwFLGK+yRR6tf7NngHpJPfBrz/z22rV+b+v38+g2nawaVfmks4xxcXQ9+Em9PJlMa901AQtUJHSI4oRGmcQvxCuBHt1KwqKwkVqqGykNwxIQnBT0Ba+TGlT6ciUZL8Frktw2Tq4qH6FdZ1N+TpGi2ggq++32q4sk9yRZ3gB52/U27p2wMYObbTYFeW9cz/8Xoa4TFmz1udqwTHQG1YKP72s063ZotmHWIzQbPlDZVYN0W55tqPpjka7fJ13ng5HujbGNcz33baRrGpr5hqHLke7iZAb8iWbbO1IEnOZj5jtmvtfPfF0jb2nOH7uZ7xCEL5H5+iexcEc5LFQMWHAExAiInwUI5xxALK50PFIvQ2Pq8q7UZfHxUxfHmU/1czjPfeOO0e2Zcs1T6sulLwj1IrR5bhBiQVacwG848vTI0yd4OsRFXAYDutL5nTNA2tc6v0P9J4caJD6zkG7oiIkRE78WE0PPea6Gif4zzoEnzs35WnMsGABmArmEqEO75rRt3amjPR1fZiGUh6ZWNE1pxsz7Zz/+fltnrAfMixjctq1H/IS3LKEvOCTx25tlexjUJUx9oU49j7cN4SdB8S4HBRWSj84a62/sEVrtnN1akG1Cshk9H+uEwXj1WwEhC3ZpGcevvhlQceSnda044+yzyiqNLNiyVHas0k9g7cFsuSxXqBR1Cnma9pldBpvI1rHpDGBzMPv0L4HN/kPYZr0ihxx23cW4VI1L1a9cqvzrLVUgtu+UVRug9s085/4/</diagram></mxfile>
```
mmccool, line 2: Not sure what the heck this is, but I don't think it belongs here.  Also there are embedded quotes and angle brackets in this thing which may cause parsing issues.

index.html
```diff
@@ -656,29 +656,74 @@ <h4>Registration Information</h4>
...
+                        and returned to consumers in <a>Enriched TD</a>s to allow
+                        bookkeeping and search.
+                        <span class="rfc2119-assertion" id="tdd-registrationinfo-created-updated">
+                            The maintenance of creation and modification time of <a>TD</a>s inside
+                            the directory are OPTIONAL for servers.
+                        </span>
+                    </p>
+                    
+                    <p>
+                        The `expires` attribute is set by the producer and submitted to the directory
```
mmccool, line 694: The directory may also have a maximum value that it may be set to, and may provide a default.  So really the registration response should be the final result: the producer "asks" for a certain expiry, the directory either accepts it or uses a possibly shorter value, then returns the result (e.g. in an enriched TD).   The directory should also be able to set a finite expiry value even if the requester does not specify (and then it is (always) the responsibility of the the requester to re-register to update the expiry before the TD expires).

In other words, TTL should always be finite...

index.html
```diff
@@ -656,29 +656,74 @@ <h4>Registration Information</h4>
...
+                        
+                        The expiry is also a useful indicator to inform the consumers about
+                        expiry of dynamic <a>TD</a>s, e.g. when changes to metadata such as
+                        geolocation or properties is expected to be valid for a limited period.
+                        
+                        Lastly, consumers may rely on the expiry date
+                        (in a similar fashion to 
+                        <a href="https://tools.ietf.org/html/rfc7234#section-5.3">Expires</a> HTTP header)
+                        to know how long a retrieved <a>TD</a> will be valid
+                        and when they need to request for a latest one.
```
mmccool, line 724: "request a more recent one."

mmccool:
OK, but I think we need to make setting the expiry a negotiation between the client and the directory.  A directory implementation may have a policy to only store TD for a certain maximum length of time.  If a client requests an expiry longer than that, it should not get it, but the shorter period.  If a client does not explicitly request an expiry, the directory may still want to set one (and again, the client needs to be informed and needs to deal with it, i.e. updating the TD as needed).

A separate issue we should maybe think about (not necessary to include in this PR) is whether the directory should "poll" Things for updated TDs, i.e. as an option during registration, a Thing might register a callback URL that indicates where to get new TDs and at what interval to retrieve them; this polling interval would have to be less than the expiry interval.  This would eliminate the client's responsibility to re-register before the expiry.  Complications with this would be client security, etc.

index.html
farshidtz: I agree that making registrations time-limited is useful in most scenarios, however, making it mandatory while not prescribing a maximum value defeats its purpose. Servers could set it to e.g. 10 years by default which is useless. On the other hand, we can't decide on a maximum value that is appropriate for all applications.

mmccool:
I added an issue about polling so we can discuss that possible extension there: https://github.com/w3c/wot-discovery/issues/164

Along the way I was wondering whether expiry should be an absolute dateTime (and it should be a dateTime, not just a date) or relative to the last update dateTime (e.g. a time-to-live interval).

mmccool:
Anyhow, I see "expires" is defined as a xsd:datetime in the ontology.

farshidtz:
> Along the way however I was wondering whether expiry should be an absolute dateTime (and it should be a dateTime, not just a date) or relative to the last update dateTime (e.g. a time-to-live interval).

I thought about having it as a relative TTL value, but I had two reasons against it (note that our implementation currently uses TTL and calculates expiry):
- It doesn't work well with every protocol. E.g. a registration may be sent to an MQTT broker and reach the directory few hours later when the directory goes online. 
- TTL may be confusing if compared with caching mechanisms (e.g. in DNS). Should clients / proxies assume the retrieved resource is valid for TTL amount of time from the retrieval time? Or will it be digested by directory and replaced with expires?

context/discovery-ontology-chowlk.xml
mmccool: never mind the syntax issue, still would rather see an SVG here or something standard.  But can live with it for now.  Not sure it is even used?   Is it just for bookkeeping by the tool?

mmccool:
Discussion
- relative vs absolute: relative actually useful if registration delayed, avoids having things expire before they can be stored; also issues with clock skew; also, what if Things don't have clocks?
- return expiry in enriched TD should always be absolute...
- but request (during registration) could be absolute or relative expiry; always stored as an absolute time relative to the local clock; should the directory also embed the time of retrieval in the TD?   Could look at the header (which has the dateTime in HTTP) but this is protocol-specific
- Could have "time remaining until expiry" calculated by the directory and included in the TD; could be simple seconds (with possible fractional part).

Use case:
1. Thing with no clock, needs to periodically update its TD before it expires.  How does it do this?
2. Delays due to proxies or brokers

farshidtz:
> 2. Delays due to proxies or brokers

I am skeptical about the usefulness of relative expiry time to make up for delays. If the delay is predictable, the client can set an absolute expiry that accounts for it. If it is unpredictable, a relative expiry does not help and may result in registration of a TD for a device that is no longer online.

I would propose a `ttl` attribute which indicates how many seconds the TD is considered valid from the moment of interaction with the server. The following are necessary for devices without a clock:

1. During registration, if a `ttl` is present, the directory would calculate `expires` time as `t+ttl`.
2. During retrieval, `expires` will show the expiry time and `ttl` will be set to `expires-t`.

where `t` is the current server time.

mmccool:
So if I understand this, upon registration a Thing can either set "expires" (absolute) or "ttl" (relative), but not both.  When a TD is retrieved, both will be provided (and the ttl basically lets you know how long before a Thing has to renew).  Would the TTL be provided as a single number, e.g. seconds, with a fractional part?

Anyway, if the above is all correct then I'm good with the proposal.  I think if a device without a clock can use ttl as a simple number of seconds (only) it will simplify registration for such devices.

farshidtz:
@mmccool yes, correct.

But I found a flaw in that design: If the producer sets the `ttl` before signing and the directory resets it to the remaining life time at retrieval, the signature would no longer be verifiable.

For example, the following registration with `ttl` of 100 seconds:
```
{
    "@context": "http://www.w3.org/ns/td",
    "id": "urn:example:shortlived",
    "registration": {
        "ttl": 100
    },
    "security": "nosec_sc",
    "securityDefinitions": {
        "nosec_sc": {
            "scheme": "nosec"
        }
    },
    "title": "TD With TTL"
}
```

Will be retrieved with a different `ttl`:
```
GET /td/urn:example:shortlived HTTP/1.1
Host: localhost

HTTP/1.1 200 OK
Content-Type: application/ld+json
Vary: Origin
Date: Thu, 29 Apr 2021 09:12:07 GMT
Content-Length: 313

{
    "@context":"http://www.w3.org/ns/td",
    "id":"urn:example:shortlived",
    "registration":{
        "created":"2021-04-29T07:29:51.193185Z",
        "modified":"2021-04-29T09:11:39.866041Z",
        "expires":"2021-04-29T09:13:19.866041Z",
        "ttl":72
    },
    "security":"nosec_sc",
    "securityDefinitions":{
        "nosec_sc":{
            "scheme":"nosec"
        }
    },
    "title":"TD With TTL"
}
```

farshidtz:
I made some changes:
- Assertion to recommend setting an upper limit and refusing incompiant registrations
- TTL for setting relative expiry by producer
- Retrieved Date for processing of expiry and other timestamps by clock-less clients. This is mostly useful for protocols other than HTTP.

context/discovery-ontology-chowlk.xml
farshidtz: @AndreaCimminoArriaga any idea?

AndreaCimminoArriaga: Hi, is the file that generates the tool that I used to generate the ontology diagram. It is like the plantUML files.

mmccool:
Thoughts:
1. If signing can pick and choose what can be signed (select a subset), then can omit TTL.
2. Still a problem if client sets TTL.  Current version leaves TTL alone, though.   Since TTL is fixed in the revision, it can be part of the signature.

index.html
```diff
@@ -651,35 +651,128 @@ <h4>Registration Information</h4>
...
                             </tr>
-                            <!-- <tr class="rfc2119-table-assertion" id="tdd-vocab-expires--Thing">
+                            <tr class="rfc2119-table-assertion" id="discovery-vocab-expires--RegistrationInformation">
                                 <td><code>expires</code></td>
-                                <td>Provides information when the TD instance registration will
-                                    expire.</td>
+                                <td>
+                                    Provides the time for when the TD instance registration expires.
+                                    <p>
+                                        The producer MAY set this to indicate the absolute expiry time
```
mmccool, line 691: be careful to avoid implying that there is a different data type for absolute and relative expiry time.  Rather than using these terms I would explicitly refer to "ttl" and "expiry" (e.g. say if one is present, the other should not be, etc).

index.html
```diff
@@ -651,35 +651,128 @@ <h4>Registration Information</h4>
...
                                 <td>optional</td>
                                 <td><a href="http://www.w3.org/TR/2012/REC-xmlschema11-2-20120405/#dateTime"><code>dateTime</code></a></td>
-                            </tr> -->
+                            </tr>
+                            <tr class="rfc2119-table-assertion" id="discovery-vocab-ttl--RegistrationInformation">
+                                <td><code>ttl</code></td>
+                                <td>
+                                    Time-to-live: relative amount of time in seconds from the registration time 
+                                    until when the TD instance registration expires. 
+                                    This attribute is ignored if there is an absolute expiry time.
```
mmccool, line 708: "is an absolute expiry time" -> "if <code>expires</code> is present", etc.

index.html
```diff
@@ -651,35 +651,128 @@ <h4>Registration Information</h4>
                             <tr>
                                 <th>Vocabulary term</th>
                                 <th>Description</th>
-                                <th>Assignment</th>
+                                <th>Client Assignment</th>
```
mmccool, line 654: not clear if client is "Registrant" or "Querier/Consumer".  I would add another column and distinguish these two.  A Registrant "sets" the ttl, for example, but a Consumer can only read it from the Directory.  Also, Server should probably be "Directory".

PR #163, opened 2021-04-26
Add pagination references and improve anonymous TD definition

farshidtz:
See individual commits addressing the feedback from Discovery call on April 19th.

***
<a href="https://pr-preview.s3.amazonaws.com/farshidtz/wot-discovery/pull/163.html" title="Last updated on Apr 26, 2021, 10:40 AM UTC (d3f6ef2)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-discovery/163/581dacc...farshidtz:d3f6ef2.html" title="Last updated on Apr 26, 2021, 10:40 AM UTC (d3f6ef2)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -248,7 +248,7 @@ img.wot-diagram {
         <dl>
             <dt><dfn data-lt="WoT Anonymous Thing Description">Anonymous TD</dfn>
             </dt>
-            <dd>A Thing Description without an identifier (`id` attribute).
+            <dd>A Thing Description without a user-defined identifier (`id` attribute).
             </dd>
             <dt><dfn data-lt="WoT Discovery">Discovery</dfn>
             </dt>
@@ -638,7 +638,7 @@ img.wot-diagram {
                     the <a>Thing Description Directory</a> can contain zero or more <a>TD</a>s.
                     For every TD, the directory additionally maintains the registration
                     information for bookkeeping and search purposes.
-                    A TD which embeds the registration information is called an <a>Enriched TD</a>.
+                    <a>Enriched TD</a>s embed the registration information inside them.
                     The ontology of a <a>TD</a> maintained by the directory is illustrated
                     in [[[#directory-td-class-diagram]]].
                 </p>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -682,18 +682,24 @@ <h4>Registration Information</h4>
...
-                    begin with `_:`. For example,
-                    a TD that has local identifier `48951ff3-4019-4e67-b217-dbbf011873dc`
-                    will have the following blank node identifier: `_:48951ff3-4019-4e67-b217-dbbf011873dc`.
+                    begin with `_:` which look like IRIs with an underscore scheme. 
+                    <p>
+                        For example,
+                        an <a>Anonymous TD</a> that has local identifier `48951ff3-4019-4e67-b217-dbbf011873dc`
+                        will have the following blank node identifier
+                        set as its `id` attribute: `_:48951ff3-4019-4e67-b217-dbbf011873dc`.
+                        <a>Anonymous TD</a>s that embed blank node identifiers are in
```
mmccool, line 700: This is still a little ambiguous.  It should say the id is ONLY embedded if an enriched form is requested.  TD should not be modified from their original form unless there is an explicit request to "enrich" them.

index.html
```diff
@@ -682,18 +682,24 @@ <h4>Registration Information</h4>
...
                 <section id="exploration-directory-anonymous-td" class="normative">
                     <h4>Anonymous TD Identifiers</h4>
-                    <a>Anonymous TD</a>s have no `id` (or `@id`) and are considered as blank nodes [[JSON-LD11]].
-                    The server assigns a local identifier to such TDs to allow management and retrieval
-                    from the directory.
+                    <a>Anonymous TD</a>s are considered as JSON-LD blank nodes [[JSON-LD11]].
+                    The server assigns a local identifier to such TDs
+                    to enable management and retrieval from the directory.
                     <span class="rfc2119-assertion" id="tdd-reg-anonymous-td-identifier">
                         In situations where the server exposes an Anonymous TD, it MUST
```
mmccool, line 689: MUST means always, I thought it was to be embedded only in enriched form?  I disagree with always embedding the blank node id.

index.html
```diff
@@ -638,7 +638,7 @@ <h3>Information Model</h3>
                     the <a>Thing Description Directory</a> can contain zero or more <a>TD</a>s.
                     For every TD, the directory additionally maintains the registration
                     information for bookkeeping and search purposes.
-                    A TD which embeds the registration information is called an <a>Enriched TD</a>.
+                    <a>Enriched TD</a>s embed the registration information inside them.
```
mmccool, line 641: There could be other kinds of metadata embedded in a TD besides registration information.  I would generalize this to say "metadata, including registration information"

mmccool:
General comment: the wording says an id MUST be embedded, which I am reading as "always" but I thought it would only be included in enriched TDs.  Let's discuss.  I don't think TDs should be modified unless there is a request to do so.   Is this being motivated by a JSON-LD technical requirement?

mmccool:
Discussion from meeting:
- Can only include id in enriched TDs
- Can make enriched the default
- Still don't have the "enriched" query parameter; proposal is to make "unenriched" the option
- OR we could add an "original" option... which would return the EXACT string provided during registration (if the input was canonicalized, otherwise it returns an "unenriched" form with all the same data as the input, but order of elements, white space, etc. might be different).

PR #168, opened 2021-05-10
Add alternative payload format

wiresio:
Base for discussion about introducing an alternative payload format allowing to send pagination info in body.
- Close allignment w/ Hydra (e.g. usage of context) - yes/no?
- Description sufficient?
- References to be fixed

***
<a href="https://pr-preview.s3.amazonaws.com/wiresio/wot-discovery/pull/168.html" title="Last updated on May 10, 2021, 8:59 AM UTC (9fe0a99)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-discovery/168/13d6508...wiresio:9fe0a99.html" title="Last updated on May 10, 2021, 8:59 AM UTC (9fe0a99)">Diff</a>

### DIFF
--- a/background.md
+++ b/background.md
@@ -152,6 +152,10 @@ In contrast to the above technologies, here, a central directory can be used for
 * **Research article: "A Semantic Enhanced Service Proxy Framework for Internet of Things"**
     - A semantic based framework which uses the concept of service advertisement of a smart object. Authors argue that such a mechanism makes the service registration easier which in turn facilitates discovery. The advertisement contains a service metadata including name, id, endpoint, location and semantic annotation link.
     - Reference: {{Alam2010}}
+	
+* **Hydra Core Vocabulary: "A Vocabulary for Hypermedia-Driven Web APIs"**
+    - Hydra is a lightweight vocabulary to create hypermedia-driven Web APIs. By specifying a number of concepts commonly used in Web APIs it enables the creation of generic API clients.
+    - Reference: {{Hydra}}
 
 #### Searching Across Peers
 
--- a/index.html
+++ b/index.html
@@ -1067,6 +1067,35 @@ img.wot-diagram {
                                 A successful response MUST have 200 (OK) status, contain `application/ld+json`
                                 Content-Type header, and an array of TDs in the body.
                             </span>
+
+							<p>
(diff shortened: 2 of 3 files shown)


mmccool:
Discussion:
1. This still needs chunking/framing, so would be in addition to that.
2. Pagination can still be done simply with appropriate queries, and an example for that would be helpful.
3. Returning different content types from a single interaction will complicate the scripting usage - perhaps a different interaction would be better.  This would remove any potential issues with use of content negotiation with scripting, in different protocols, etc.
4. This complicates implementation - could it be optional?  But then consumers can't depend on it.   If it was a separate interaction that was not required, then if it was not in the TD, then a consumer would not be able to use it.

wiresio:
1. Yes, would be in addition. Chunking anyway is optional.
2. Yes, however downloading all TDs incl. pagination from the TDD should be explicitly available (via listing) and not just implicitly (via search).
3. What about adding another query parameter for this, like ?pagination=inband/outband? Content negotiation will not help since both responses have contentType as "application/ld+json".
4. Could be optional such that only one reference implementation has to support it. Consumers can depend on the "pure" listing API and might access an optional "convenience" listing API.

farshidtz:
> What about adding another query parameter for this, like ?pagination=inband/outband? Content negotiation will not help since both responses have contentType as "application/ld+json".

The [hydra](http://www.hydra-cg.com/spec/latest/core/) spec does use `application/ld+json` in examples but it doesn't recommend or mandate it as far as I looked. The hydra spec itself is an unofficial draft. I think something like `application/vnd.hydra+json` would work.

Alternatively, similar to what you suggested, adding a query parameter such as `?hydraCollection[=true]` could be used to override the default. Or `?collectionFormat=array|hydra` if we want to make it extensible.

mmccool:
Discussion (12 July):
- Needs to be updated to match latest spec
- No *strong* objections, especially if it is just another way of doing things
- Regarding mandatory vs. optional; still need two implementations even for optional features; however, consensus is to make it optional
- Need to clarify when to use the "envelope" vs. using "listing" (perhaps some text to explain when each is appropriate; can be added to this PR).

mmccool:
Replaced with new PR, will close this one without merging.

PR #170, opened 2021-05-10
Ignoring unknown event types

farshidtz:
As pointed in https://github.com/w3c/wot-discovery/pull/159#issuecomment-832111684, adding new event types may break existing clients. This PR adds a statement to enable event type extensions without breaking existing clients.

***
<a href="https://pr-preview.s3.amazonaws.com/farshidtz/wot-discovery/pull/170.html" title="Last updated on May 10, 2021, 2:10 PM UTC (4f678dc)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-discovery/170/13d6508...farshidtz:4f678dc.html" title="Last updated on May 10, 2021, 2:10 PM UTC (4f678dc)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1325,6 +1325,7 @@ img.wot-diagram {
                                 the Thing Descriptions within the directory using 
                                 `create`, `update`, and `delete` event types.
                             </span>
+                            Clients should ignore any event with a type that is unknown to them.
                         </dd>
 
                         <dt>Event Filtering</dt>


index.html
```diff
@@ -1325,6 +1325,7 @@ <h4>Notification</h4>
                                 the Thing Descriptions within the directory using 
                                 `create`, `update`, and `delete` event types.
                             </span>
+                            Clients should ignore any event with a type that is unknown to them.
```
mmccool, line 1328: - Shouldn't this be part of the TD spec?
- Shouldn't it be an RFC2119 assertion?
- How do we test it?

mmccool:
Conclusion: valid point, but probably should (a) be in the TD spec, as it applies to all Things (b) be an assertion.
Actions: close this PR without merging, create an issue in the TD spec.  TD issue https://github.com/w3c/wot-thing-description/issues/1134

Issue #143, opened 2021-03-17
SHACL validation


mmccool:
- when and how SHACL validation should be done?
- error report
- extension validation

AndreaCimminoArriaga:
SHACL should be used after registering a TD. The TD must be translated into any RDF representation, and then, the SHACL applied. Consider that SHACL shape must contain restrictions for all the ontologies used in the document, not only the TD. As I mentioned in [a previous issue](https://github.com/w3c/wot-thing-description/issues/1015), the service [Astrea](https://astrea.linkeddata.es/) allows to generate quite rich SHACL shapes automatically from a given ontology. It could be used for generating a first version of the TD shape, however, consider that if someone uses the TD with SAREF that shape will not be enough (although it will still validate the TD part).

vcharpenay:
> for generating a first version of the TD shape

There already is [a SHACL shape for TD documents](https://github.com/w3c/wot-thing-description/blob/main/validation/td-validation.ttl). It is in fact the main source for rendering the TD spec (see [`templates.sparql`](https://github.com/w3c/wot-thing-description/blob/main/templates.sparql)).

AndreaCimminoArriaga:
> There already is a SHACL shape for TD documents. It is in fact the main source for rendering the TD spec (see templates.sparql).

That is great I will start using it in [my library for translating](https://github.com/oeg-upm/wot-jtd) TDs from JSON-LD 1.1 framed to RDF and back.

> An SHACL definition that would be automatically generated from Astrea couldn't be complete, by the way. It would miss e.g. default values and a higher-level constraint that at least one security configuration must be provided.

That is true, but in the assumption there was no shape the one generated by Astrea could be a good start. Even if the shape may not include default values, all the structural restrictions defined in the ontology would be included (like the security configuration). Anyway having already a shape is a very good new.

mmccool:
I am trying to add a formal definition of validation for inclusion in the TD spec.  Please see https://github.com/w3c/wot-thing-description/pull/1085
Note that my definition of "Intermediate" validation includes SHACL validation and would be appropriate for directories.

mmccool:
- SHACL can be expensive (needs RDF processing...), few implementations
- Does catch more things than JSON Schema
Conclusion:
- MUST: Minimal validation (JSON Schema).  Validation of directory and enrichment extensions is mandatory (MUST).  Validation of other extensions (iot, qudt, etc). is optional (MAY).
[...]

PR #180, opened 2021-05-20
Create JSON Schema for validation of discovery extensions

farshidtz:
This may be used to validate the discovery extensions in combination to other schemas.

The JSON Schema for validation of native TD attributes: https://github.com/w3c/wot-thing-description/blob/main/validation/td-json-schema-validation.json
 
I used the following filename template to allow adding other formats (e.g. SHACL #143) and file extensions (e.g. ttl) in future:
```
/validation/td-discovery-extensions-{format}.{ext}
```

### DIFF
--- a/validation/td-discovery-extensions-json-schema.json
+++ b/validation/td-discovery-extensions-json-schema.json
@@ -1,0 +1,32 @@
+{
+  "title": "WoT Discovery TD-extensions Schema - 21 May 2021",
+  "description": "JSON Schema for validating TD instances with WoT Discovery extensions",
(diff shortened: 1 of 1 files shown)


AndreaCimminoArriaga (approved):
I think the schema is ok, but maybe it has some things missing depending on how the PR #175 is finally solved. The PR is ready to be merged, and aligned with the current model but is likely to be changed after the resolution of PR #175

AndreaCimminoArriaga:
I have prepared a SHACL file to add to this `/validation` folder that can be used to validate the discovery extension, when this PR get's merged I will create a new one with the SHACL

PR #213, opened 2021-07-18
Add alternative payload format (2)

### LINKED ISSUE #16: Handle huge set of Thing Descriptions (pagination, streaming, etc.)

wiresio:
Rework of https://github.com/w3c/wot-discovery/pull/168

resolves #16

***
<a href="https://pr-preview.s3.amazonaws.com/wiresio/wot-discovery/pull/213.html" title="Last updated on Jul 26, 2021, 3:52 PM UTC (034d81f)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-discovery/213/6f0f355...wiresio:034d81f.html" title="Last updated on Jul 26, 2021, 3:52 PM UTC (034d81f)">Diff</a>

### DIFF
--- a/context/discovery-context.jsonld
+++ b/context/discovery-context.jsonld
@@ -9,7 +9,17 @@
       "@id": "discovery:ThingLink",
       "@type": "@id"
     },
-    "registration": {
+	"ThingCollection": {
+      "@id": "discovery:ThingCollection",
+      "@type": "@id"
+    },
+	"totalItems": "discovery:contains",
+	"member": [
+	  "@graph"
+	],
+	"@id": "discovery:thisThingCollection",
+	"next": "discovery:nextThingCollection",
+	"registration": {
       "@id": "discovery:hasRegistrationInformation",
       "@type": "@id",
       "@context": {
--- a/context/discovery-context.jsonld
+++ b/context/discovery-context.jsonld
@@ -13,8 +13,8 @@
       "@id": "discovery:ThingCollection",
       "@type": "@id"
     },
-	"totalItems": "discovery:contains",
-	"member": [
+	"total": "discovery:contains",
+	"members": [
 	  "@graph"
 	],
 	"@id": "discovery:thisThingCollection",
--- a/context/discovery-context.jsonld
+++ b/context/discovery-context.jsonld
@@ -15,7 +15,7 @@
     },
 	"total": "discovery:contains",
 	"members": [
-	  "@graph"
+	  "td:Thing"
 	],
 	"@id": "discovery:thisThingCollection",
 	"next": "discovery:nextThingCollection",
--- a/index.html
+++ b/index.html
@@ -1404,44 +1404,35 @@ img.wot-diagram {
 						<p>
 							For pagination information the alternative format leans against the formats as described
 							in <a href="https://www.hydra-cg.com/spec/latest/core/#advanced-concepts">
-							Hydra Advanced Concepts</a>. More concretely the
+							Hydra Advanced Concepts</a>, more concretely the
 							<a href="https://www.hydra-cg.com/spec/latest/core/#example-20-a-hydra-partialcollectionview-splits-a-collection-into-multiple-views">
-							Partial Collection View</a> using the <i>members</i> field to accomodate the array
-							of TDs is applied, and looks like as follows for the listing endpoint:
+							Partial Collection View</a>. Adapted to our purposes and using the <i>member</i> field to accomodate the array
+							of TDs, it looks like as follows for the listing endpoint:
 						</p>
 						<p>
 							<aside class="example" id="example-alternative-payload"
 								title="Alternative payload format including pagination information">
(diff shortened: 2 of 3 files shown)


farshidtz:
Thanks. I think the content is fine, but it needs some more work:

- The example is not correct. `totalItems` is 10 and all those 10 are listed: so no next pages. There is still a `nextLink` with offset 1. Should there be a `nextLink` when there are no remaining pages? If yes, the offset should be 10. I think a better fix would be to set total to 12 and `nextLink` to `/things?offset=10&limit=10` to be consistent with the other pagination examples (array format).
- The API endpoint should be `/things`, not `/td`
- Missing API spec for the new interaction. I think addition of the new uriVariable (`?format=array|hydraCollection`) on the existing pagination form would suffice. 

Some editorial issues:
- The new paragraphs should be added at the end of Listing (before errors) as this is an alternative to pure array. The existing paragraph starting with "Serializing and returning the full list of TDs may be burdensome to servers. As such, servers should serialize incrementally and utilize protocol-specific mechanisms to respond in chunks...." should be moved to after the hydra collection. 
- Missing hyrda references
- Code as italic text instead of using `<code>` tag or backquotes

wiresio:
Thanks for the feedback @farshidtz ! I applied your comments and will need some help with:

- Add references
- "code" instead of "i"
    
Let's resolve it in today's meeting.

index.html
```diff
@@ -1415,6 +1395,76 @@ <h4>Listing</h4>
...
+							send a more verbose payload allowing server-side information, like pagination
+							information, to be put in addition to the actual data.
+						</p>
+							
+						<p>
+							For pagination information the alternative format leans against the formats as described
+							in <a href="https://www.hydra-cg.com/spec/latest/core/#advanced-concepts">
+							Hydra Advanced Concepts</a>. More concretely the
+							<a href="https://www.hydra-cg.com/spec/latest/core/#example-20-a-hydra-partialcollectionview-splits-a-collection-into-multiple-views">
+							Partial Collection View</a> using the <i>members</i> field to accomodate the array
```
farshidtz, line 1409: "members" -> "member" ?

mmccool:
Needs some refinement.  In particular, as discussed, it's troublesome to depend normatively on a CG document, and it's also a problem to just define an extension point (with hydra collections as examples) since that creates interop problems.  We need a built-in normative way to do this, but can't just adopt the hydra ontology directly.  So it was suggested that we resurrect the design we had discussed in https://github.com/w3c/wot-discovery/issues/16 and which can be seen in the context history, see e.g. https://github.com/w3c/wot-discovery/commit/de02ff02bc38c3a846ca98ee9ae7ec75b27ad800#diff-94098f295e6c8f89beca5d1221bbaefaf5c410e17b1a6be3d5df9051b205f72f

farshidtz:
Since we are not strictly following hydra anymore, why don't we use plural `items` instead of the single `member`? 

https://github.com/wiresio/wot-discovery/blob/66284902a8892fdf9764900f75c51c21fe958f36/index.html#L1419-L1423
```
"totalItems": 12,
"member": [
	{ TD },
	9 other TDs...
],
```

to
```
"totalItems": 12,
"items": [
	{ TD },
	9 other TDs...
],
```

The following maps better:
totalItems -> items
totalMembers -> memberS

wiresio:
Thanks for the comment!
Actually I'd like:
_total + members_

context/discovery-context.jsonld
```diff
@@ -9,7 +9,17 @@
...
       "@type": "@id"
     },
-    "registration": {
+	"ThingCollection": {
+      "@id": "discovery:ThingCollection",
+      "@type": "@id"
+    },
+	"total": "discovery:contains",
+	"members": [
+	  "@graph"
```
mmccool, line 18: This perhaps should be td:Thing included from the td context, since in the intent is to return TDs (formatted as TDs) and not general RDF graphs.

AndreaCimminoArriaga:
I have to review in detail the new context, but maybe it would be better to have a separate context with the pagination terms, and the range of the term members should point to td:Thing or any of the types defined in the discovery context. In this way this second context is totally optional and would point to the main one

PR #221, opened 2021-09-20
Update intro to introduction mechanisms

mmccool:
Clean up intro to introduction mechanisms, to make clear that:
- Any of the introduction mechanisms can be used
- Multiple introduction mechanisms can be used
- No particular introduction mechanism is mandatory
- The purpose is first contact
- The result is always a URL (NOT TDs or other metadata)
- There might be more than one result and if so they should be used (so in general the result of the introduction process is a SET of URLs pointing to exploration mechanisms).

We might have to clean up some of the text, in particular if it says anywhere an introduction returns a TD.  The result should always be a URL, full stop.

***
<a href="https://pr-preview.s3.amazonaws.com/mmccool/wot-discovery/pull/221.html" title="Last updated on Sep 20, 2021, 2:20 PM UTC (58565d8)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-discovery/221/ef9288b...mmccool:58565d8.html" title="Last updated on Sep 20, 2021, 2:20 PM UTC (58565d8)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -306,9 +306,15 @@ img.wot-diagram {
     </section>
     <section id="introduction-mech" class="normative">
         <h1>Introduction Mechanisms</h1>
-        <p>This chapter describes a mechanism for discovering a Thing or a <a>Thing Description Directory</a>.
-           The following mechanism is provided by the Thing or the <a>Thing Description Directory</a> so that
-           Consumers can discover the Thing Description or a URL that point to the Thing Description.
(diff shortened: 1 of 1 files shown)


mmccool:
reviewed; there are indeed a few issues in direct and well-known that talk about fetching TDs that should be cleaned up, but we can do that in a future PR

Issue #156, opened 2021-04-19
JSONPath and XPath response data models


farshidtz:
The current spec has the following assertion for both JSONPath and XPath responses:
> ... in the body a set of complete TDs or a set of TD fragments.

We need to check if this is valid.

What happens when a single attribute is selected? What about functions (e.g. count)?

Raised in #153

AndreaCimminoArriaga:
Would not be a string (single attribute) just a TD fragment? is still a part of a major TD, why should it not be a TD fragment?
Are you proposing to define a kind of json body wrapping the search results?

farshidtz:
It may not be **a set** of TD fragments. The result could be:
```json
5
```
for counting or
```json
"urn:example:some-td"
```
for querying just an id.

Note the the above are valid JSONs.

farshidtz:
For JSONPath, this seems to be implementation-specific right now. 

Picking a single object/element is not always returned as an array: See Nebhale: https://jsonpath.herokuapp.com/
Or as mentioned above, some aggregator functions will not give arrays.

But the IETF draft suggests returning list always so our text would be correct: 
https://datatracker.ietf.org/doc/html/draft-ietf-jsonpath-base-00#section-3.5

mmccool:
Note there is a new version now: https://datatracker.ietf.org/doc/html/draft-ietf-jsonpath-base-01

mmccool:
Problem is, JSONPath is currently a moving target and the IETF RFC is unlikely to be finalized prior to our (original) target date for a Discovery PR.  So we have a couple of choices:
1. Wait on our REC until JSONPath IETF RFC is complete
2. Make XPath the "primary" query mechanism, make JSONPath optional/experimental
      - In 1.1 can adopt the JSONPath standard
      - XPath could be optional or mandatory
      - It would be good to have at least on mandatory query type
      - Right now only have one XPath implementation
3. Do nothing, leave JSONPath as primary and mandatory
      - but... can't depend on draft IETF spec
      - ... so not really an option

mmccool:
4. Make all query types optional.
5. Write JSONPath spec inside Discovery spec.  But: large, may drift from "official" spec to come.
6. Like 5, but allow only a small subset of simple queries for targeted use cases, picking things that are unlikely to change.  Ex: keyword search, search based on @type.  Look at what we did for Node-RED, for instance... (@relu91 has volunteered to look into this).

mmccool:
[...]

PR #223, opened 2021-09-20
Add object types to actions

### LINKED ISSUE #182: TDD TD: Don't Actions need an input field
The TDD TD has the following actions: `createTD`, `updateTD`, `updatePartialTD`. Don't these actions need an input that has an object type, which is the TD ?

farshidtz:
Closes #182

I also added object types for search actions. But we need to clarify JSONPath and XPath models later on: https://github.com/w3c/wot-discovery/issues/156

***
<a href="https://pr-preview.s3.amazonaws.com/farshidtz/wot-discovery/pull/223.html" title="Last updated on Oct 1, 2021, 3:08 PM UTC (d5a3099)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-discovery/223/dad5c4f...farshidtz:d5a3099.html" title="Last updated on Oct 1, 2021, 3:08 PM UTC (d5a3099)">Diff</a>

### DIFF
--- a/directory.td.json
+++ b/directory.td.json
@@ -137,6 +137,7 @@
                 }
             },
             "input": {
+                "description": "The schema is implied by the content type",
                 "type": "object"
             },
             "forms": [
@@ -162,6 +163,7 @@
         "createAnonymousThing": {
             "description": "Create an anonymous Thing Description",
             "input": {
+                "description": "The schema is implied by the content type",
                 "type": "object"
             },
             "forms": [
@@ -201,6 +203,7 @@
                 }
             },
             "output": {
+                "description": "The schema is implied by the content type",
                 "type": "object"
             },
             "safe": true,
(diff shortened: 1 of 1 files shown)


mmccool:
- Output is always given the type "object", but JSON Path queries might result in a fragment, e.g. a string, number, etc.  Maybe we should use "oneOf"?   Text currently says "object", however.  See #156; this is not clear in the JSON Path spec, but some implementations do permit fragments as results
- No data schema, just a type.  Is that legal, or does it describe an empty object?   OK if schema is optional (it is) and if additional properties are allowed (they are).
- We really should add some explanatory text somewhere, saying (for example) the schema is implied by the content type (for actions that take a TD as input) or the query (for actions that generate output as the result of a JSON path or XPath query).

mmccool:
For the last point above, I would add the following sentence to the end of the first paragraph in 6.2.2.1.1 (for example):
"Note that in the TD describing the input schema for this action no schema is given, because it is implied by the content type."

Elsewhere this might be
"Note that in the TD describing the output schema for this action no schema is given, because it is implied by the query."

This could also go in a "description" field in the TD.

Issue #240, opened 2021-11-23
Move common terminology to the Architecture spec


mlagally:
There are some terminology entries that are common and should be moved to the architecture terminology section.
Examples include "Anonymous TD", Thing Description Directory, Partial TD, Discovery

k-toumura:
related issue: w3c/wot-architecture#613

mmccool:
Note that there are NO definitions in Discovery any more, the PR above moves *everything* to Architecture (well, it deletes them in Discovery, another PR to update Architecture is pending)

Issue #216, opened 2021-07-26
Define role of "Discoverer/Registerer"


mmccool:
For brownfield devices we probably need to define a "Discoverer" process/function  that generates the TD and registers it with a Directory service, or replies on the Things's behalf (for Direct discovery).   Latter is related to shadows/proxies.

To do: needs a better/less confusing name for this role...

_Originally posted by @mmccool in https://github.com/w3c/wot-discovery/issues/53#issuecomment-700663727_

farshidtz:
I just got the chance to look at this new term added by #222 .

Why is it called Discoverer instead of [Registerer](https://www.merriam-webster.com/dictionary/registerer)? This entity doesn't do discovery, it does registration.

benfrancis:
Some unsolicited feedback...

The entity which generates/produces a Thing Description for a (non-self-describing) device and the entity which registers that Thing Description with a directory may not always be the same. The former is likely to be a server (e.g. serving a Thing Description over HTTP) and the latter is likely to be a client (consuming the Directory Service API).

An entity which performs both functions is probably a "gateway" which is generating a Thing Description for a non-WoT device and exposing it via its own directory [1]. I can't otherwise imagine a Web Thing which would register itself with a directory. 

It may help to explain some concrete use cases you have in mind.

----
1.In this use case it's probably not always possible to expose all features of the Directory Service API (e.g. updating) to all clients, see #208.

mmccool:
- is a distinction between the term and the definition.  Let's get the definition right first
- regarding ben's point, the "discoverer/registerer" would fetch the TD from the generator (the server in Ben's example) and then "register" it with the TDD (as a client).  So it's still a "registerer".  But maybe we also want to define a "generator" and note that the "registerer" may or may not be the "generator".  
- definition currently talks only about registration, not fetching.  If we have both, "discoverer" may be better (implies both actions, but then definition needs updating... it was my original intention to include both).

benfrancis:
@mmccool wrote:
> regarding ben's point, the "discoverer/registerer" would fetch the TD from the generator (the server in Ben's example) and then "register" it with the TDD (as a client). So it's still a "registerer". But maybe we also want to define a "generator" and note that the "registerer" may or may not be the "generator".
> definition currently talks only about registration, not fetching. If we have both, "discoverer" may be better (implies both actions, but then definition needs updating... it was my original intention to include both).

[...]

PR #241, opened 2021-11-25
Delete common terms in terminology section

k-toumura:
(Fixed for w3c/wot-architecture#613, #240):
Move following terms from Discovery spec to Architecture spec.
- Anonymous TD
- Discoverer
- Exploration
- Introduction

And, delete following terms which are already defined in Architecture spec.
- Discovery
- Thing Description Directory(TDD)
- Partial TD

***
<a href="https://pr-preview.s3.amazonaws.com/k-toumura/wot-discovery/pull/241.html" title="Last updated on Nov 30, 2021, 6:54 AM UTC (884dab4)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-discovery/241/1b9a9ae...k-toumura:884dab4.html" title="Last updated on Nov 30, 2021, 6:54 AM UTC (884dab4)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -242,21 +242,11 @@ img.wot-diagram {
             <dfn data-lt="WoT Introduction">Introduction</dfn>,
             <dfn data-lt="WoT Thing Description Directory">Thing Description Directory</dfn>
             (<dfn data-lt="WoT TDD">TDD</dfn>),
-            <dfn data-lt="WoT Partial Thing Description">Partial TD</dfn>
+            <dfn data-lt="WoT Partial Thing Description">Partial TD</dfn>,
+            <dfn data-lt="WoT Enriched Thing Description">Enriched TD</dfn>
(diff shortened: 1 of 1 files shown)


mmccool:
There is now only one term, Enriched TD, in the Discovery spec.  Decided in Nov 29 call: move it to Architecture too and keep all Terminology centralized.

mmccool:
Also there is issue #216 which proposes a different term for "Discoverer", since it is often used by entities that KNOW about things in advance, and don't actually discover them.  So "Registerer" is more appropriate (although a nicer word would be... nice).

Decision: don't change this PR, deal with new term for Discoverer separately.

k-toumura:
> There is now only one term, Enriched TD, in the Discovery spec. Decided in Nov 29 call: move it to Architecture too and keep all Terminology centralized.

Fixed PR.

PR #243, opened 2021-12-06
Add slight wording changes

wiresio:
***
<a href="https://pr-preview.s3.amazonaws.com/wiresio/wot-discovery/pull/243.html" title="Last updated on Dec 13, 2021, 3:19 PM UTC (8ddcf77)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-discovery/243/d33a840...wiresio:8ddcf77.html" title="Last updated on Dec 13, 2021, 3:19 PM UTC (8ddcf77)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1339,7 +1339,7 @@ img.wot-diagram {
                                     </span>
                                     <span class="rfc2119-assertion" id="tdd-reg-list-pagination-header-nextlink-attr">
                                         The `next` link MUST include all arguments needed to produce the same set of
-					data and its ordering. Particularly, it MUST include the same `limit` argument
+					data and its ordering, in particular the same `limit` argument
 					given on the initial request as well as a zero-based `offset` argument anchored
 					at the beginning of the next subset.
                                     </span>
--- a/index.html
+++ b/index.html
@@ -1338,11 +1338,12 @@ img.wot-diagram {
                                         with the URL of the next subset.
                                     </span>
                                     <span class="rfc2119-assertion" id="tdd-reg-list-pagination-header-nextlink-attr">
-                                        The `next` link MUST have the same `limit` argument given on the initial
-                                        request as well as a zero-based `offset` argument anchored at
-                                        the beginning of the next subset.
+                                        The `next` link MUST include all arguments needed to produce the same set of
+					data and its ordering. Particularly, I MUST include the same `limit` argument
+					given on the initial request as well as a zero-based `offset` argument anchored
+					at the beginning of the next subset.
                                     </span>
-                                    The link may be absolute or relative to directory API's base URL.
+                                    The link MAY be absolute or relative to directory API's base URL.
                                     Moreover, it may include additional arguments that are necessary for
                                     ordering or session management.
                                 </li>
@@ -1447,7 +1448,7 @@ img.wot-diagram {
 						</p>
 							
 						<p>
-							For pagination information the alternative format leans against the formats as described
+							For pagination information the alternative format leans on the formats as described
 							in <a href="https://www.hydra-cg.com/spec/latest/core/#advanced-concepts">
 							Hydra Advanced Concepts</a>, more concretely the
 							<a href="https://www.hydra-cg.com/spec/latest/core/#example-20-a-hydra-partialcollectionview-splits-a-collection-into-multiple-views">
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -1338,11 +1338,12 @@ <h4>Listing</h4>
...
-                                        The `next` link MUST have the same `limit` argument given on the initial
-                                        request as well as a zero-based `offset` argument anchored at
-                                        the beginning of the next subset.
+                                        The `next` link MUST include all arguments needed to produce the same set of
+					data and its ordering. Particularly, it MUST include the same `limit` argument
+					given on the initial request as well as a zero-based `offset` argument anchored
+					at the beginning of the next subset.
                                     </span>
-                                    The link may be absolute or relative to directory API's base URL.
+                                    The link MAY be absolute or relative to directory API's base URL.
```
farshidtz, line 1346: To make this an assertion, it should be added to an assertion span, e.g.:
```
<span class="rfc2119-assertion" id="tdd-reg-list-pagination-header-nextlink-base">
This link MAY be ...
</snap>
```

But I suggest changing it to a MUST. IMO, "MAY" followed by "or" doesn't really prescribe anything formally.

mmccool: I agree with Farshid's rewording, using MUST.  Also, need the span tag (with a unique id, please).

index.html
```diff
@@ -1447,7 +1448,7 @@ <h4>Listing</h4>
 						</p>
 							
 						<p>
-							For pagination information the alternative format leans against the formats as described
+							For pagination information the alternative format leans on the formats as described
```
mmccool, line 1451: "leans on" --> "depends upon" or "based upon".  However "based" is not quite right, these formats are "derived from" Hydra.  My preference would be "derived from" (and "described in").

mmccool (changes requested):
after these changes are made, as discussed in Discovery call Dec 6, will merge PR.

wiresio:
@mmccool - Included the comment from yesterday. Please feel free to merge.

index.html
```diff
@@ -1338,11 +1338,14 @@ <h4>Listing</h4>
                                         with the URL of the next subset.
                                     </span>
                                     <span class="rfc2119-assertion" id="tdd-reg-list-pagination-header-nextlink-attr">
-                                        The `next` link MUST have the same `limit` argument given on the initial
-                                        request as well as a zero-based `offset` argument anchored at
-                                        the beginning of the next subset.
+                                        The `next` link MUST include all arguments needed to produce the same set of
+					data and its ordering. Particularly, it MUST include the same `limit` argument
+					given on the initial request as well as a zero-based `offset` argument anchored
+					at the beginning of the next subset.
```
farshidtz, line 1344: I believe two assertions can't be combined. Should either separate into two or:
```suggestion
                                        The `next` link MUST include all arguments needed to produce the same set of
					data and its ordering, in particular the same `limit` argument
					given on the initial request as well as a zero-based `offset` argument anchored
					at the beginning of the next subset.
```

index.html
```diff
@@ -1447,7 +1450,7 @@ <h4>Listing</h4>
 						</p>
 							
 						<p>
-							For pagination information the alternative format leans against the formats as described
+							For pagination information the alternative format is derived from the formats as described
```
farshidtz, line 1453:
```suggestion
							The alternative pagination format is derived from
```

index.html
```diff
@@ -1447,7 +1450,7 @@ <h4>Listing</h4>
...
 							in <a href="https://www.hydra-cg.com/spec/latest/core/#advanced-concepts">
```
farshidtz, line 1454:
```suggestion
							<a href="https://www.hydra-cg.com/spec/latest/core/#advanced-concepts">
```

farshidtz (changes requested):
Thanks @wiresio. There are a couple more issues. Please see my comments inline. You can probably commit the suggestions directly.

index.html
mmccool: I agree that in general combining assertions is not a good idea.  At the very least, it complicates testing.   At worst, it is ambiguous.  Also, include only one RFC2119 keyword per span (otherwise it complicates tooling) and give each assertion a unique id.

index.html
mmccool: editorial, makes sense

mmccool (approved):

PR #251, opened 2021-12-19
Refactor event affordance and event type naming

### LINKED ISSUE #197: Improving event type values in the Notification API
The [Notification API](https://w3c.github.io/wot-discovery/#exploration-directory-api-notification) supports three types of events. 

The event types has specific names used in URI variables. These were changed recently and a pending proposal suggests yet another change. But before doing so, it is better to discuss pros and cons of every naming convention.

#### Convention A
[WD1](https://www.w3.org/TR/2020/WD-wot-discovery-20201124/#exploration-directory-api-notification) used the following event types: `created_td`,  `updated_td`, `deleted_td`

Note: This is before changing `td` to `thing`/`things` across APIs. In alignment to the current spec, those types would be:  `created_things`,  `updated_things`, `deleted_things`

#### Convention B (current)
The types were recently changed in response to feedback in F2F asking for simplifications (#155). The new types are similar to `add` and `remove` from the [SSE spec](https://www.w3.org/TR/eventsource/#server-sent-events-intro) examples: `create`, `update`, `delete`

Concerns:
- It doesn't convey any information about the payload. Since the payload of specific events in relevant protocols (SSE, WS, MQTT, CoAP observer) doesn't specify the content type, having type info in the event name is helpful. Adding `thing` prefix/suffix as in conventions A and C resolves that. On the other hand, we could assume that default events always contain TDs and add prefixes/suffixes for new possible content types.
- They are action verbs which sound strange in URIs and MQTT topics. 

#### Convention C
One new proposal (#159 superseded by #184) suggests: `thing_created`, `thing_updated`, `thing_deleted`

This is an improvement to B but past tense verbs as prefixes is not so common and less readable (subscribing to thing created events?).

---
Taking a look at MQTT topic best practices and the event examples in TD spec, it looks like nouns are a more common choice. E.g. temperature, stats, status

That takes us to conventions D and E:

#### Convention D
`thing_creation`, `thing_update`, `thing_deletion`

This is very readable: subscribe to thing creation events:
`/events?type=thing_creation` (SSE):

Naively mapping it to an MQTT topic: `things/creation/{id}`

#### Convention E
Same as D but without the prefix: `creation`, `update`, `deletion` 

---

For future extensions, I can think of:
- thing expiration
- thing movement
- directory status
- directory error

### LINKED ISSUE #258: Change event affordance names
To avoid confusion, I propose changing the names of the event affordances in the Directory Service API to match the URLs proposed in #251. 

- `thingCreation` -> `thingCreated`
- `thingUpdate` -> `thingUpdated`
- `thingDeletion` -> `thingDeleted`

See #251 for further discussion.

farshidtz:
This PR changes the event type's naming convention from Convention B (e.g. `create`) to Conversion C (e.g. `thing_created`); see #197.

Adding the data type is to allow extensions to the notification API to deliver event data other than just TDs.
Changing from action verb to past verb is to make the event type more meaningful when associated with delivered events, since the type value is also to identify the type of event data on the stream.

It closes #197

---

In addition, it changes the event affordance names from nouns to past verbs for consistency with event type values. 

It closes #258

***
<a href="https://pr-preview.s3.amazonaws.com/farshidtz/wot-discovery/pull/251.html" title="Last updated on Jan 24, 2022, 2:11 PM UTC (5b525ba)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-discovery/251/eba3191...farshidtz:5b525ba.html" title="Last updated on Jan 24, 2022, 2:11 PM UTC (5b525ba)">Diff</a>

### DIFF
--- a/directory.tm.json
+++ b/directory.tm.json
@@ -414,7 +414,7 @@
             "forms": [
                 {
                     "op": "subscribeevent",
-                    "href": "/events/create{?diff}",
+                    "href": "/events/thing_created{?diff}",
                     "subprotocol": "sse",
                     "contentType": "text/event-stream",
                     "htv:headers": [
@@ -442,7 +442,7 @@
             "forms": [
                 {
                     "op": "subscribeevent",
-                    "href": "/events/update{?diff}",
+                    "href": "/events/thing_updated{?diff}",
                     "subprotocol": "sse",
                     "contentType": "text/event-stream",
                     "htv:headers": [
@@ -463,7 +463,7 @@
             "forms": [
                 {
                     "op": "subscribeevent",
-                    "href": "/events/delete",
+                    "href": "/events/thing_deleted",
                     "subprotocol": "sse",
                     "contentType": "text/event-stream",
                     "htv:headers": [
--- a/directory.tm.json
+++ b/directory.tm.json
@@ -399,7 +399,7 @@
         }
     },
     "events": {
-        "thingCreation": {
+        "thingCreated": {
             "description": "Registration of Thing Descriptions inside the directory",
             "uriVariables": {
                 "diff": {
@@ -426,7 +426,7 @@
                 }
             ]
         },
-        "thingUpdate": {
+        "thingUpdated": {
             "description": "Updates to Thing Descriptions within the directory",
             "uriVariables": {
                 "diff": {
@@ -454,7 +454,7 @@
                 }
             ]
         },
-        "thingDeletion": {
+        "thingDeleted": {
             "description": "Deletion of Thing Descriptions from the directory",
             "data": {
                 "title": "Partial TD",
--- a/index.html
+++ b/index.html
@@ -1583,7 +1583,7 @@ img.wot-diagram {
                             <span class="rfc2119-assertion" id="tdd-notification-event-types">
                                 The server MUST produce events attributed to the lifecycle of 
                                 the Thing Descriptions within the directory using 
-                                `create`, `update`, and `delete` event types.
+                                `thing_created`, `thing_updated`, and `thing_deleted` event types.
                             </span>
                         </dd>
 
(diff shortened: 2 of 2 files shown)


benfrancis (changes requested):
I really like changing the event names from "create" to "thing created" as I think that makes a lot more sense from a linguistic point of view, however I think the affordance names should be updated as well as just the URLs. The fact that the affordance name and URL use different wording is confusing.

Also note that using an underscore as a separator would mean there would be three different conventions for multi-word strings in a Directory Thing Description:
1. Lower camel case for event affordance names (e.g. `thingCreation`)
2. All lower case for operation names (e.g. `subscribeevent`)
3. Underscore separated for event URLs (e.g. `thing_created`)

I would suggest that the wording in the affordance name should at least match the wording in the URL, and ideally the case would match too, e.g. they could both be `thingCreated`.

My real preference would be to omit the URLs from the Thing Model (#253) altogether and let them be implementation-specific, but I realise we've already discussed that at some length and that is complicated by URL variables.

farshidtz:
> I really like changing the event names from "create" to "thing created" as I think that makes a lot more sense from a linguistic point of view, however I think the affordance names should be updated as well as just the URLs. The fact that the affordance name and URL use different wording is confusing.

The affordance names (thingCreation, thingUpdate, and thingDeletion) are nouns, following the TD convention for event affordances. I have no objection to changing those (to thingCreated, thingUpdated, and thingDeleted) if there is consensus, but that is unrelated to this PR.

> Also note that using an underscore as a separator would mean there would be three different conventions for multi-word strings in a Directory Thing Description:
> 
> 1. Lower camel case for event affordance names (e.g. `thingCreation`)
> 2. All lower case for operation names (e.g. `subscribeevent`)
> 3. Underscore separated for event URLs (e.g. `thing_created`)

There is already more than three different conventions in the TD, some of them out of our control.

But in URLs, we have consistently followed snake_case. E.g. `sort_by` and `sort_order`. I don't mind changing all to kebab-case. 

Other snake_case instances are in the security definitions:
https://github.com/w3c/wot-discovery/blob/c46ebf79be55ac2fb4dc4021aa29f2d4cac7e99d/directory.td.json#L12

> I would suggest that the wording in the affordance name should at least match the wording in the URL, and ideally the case would match too, e.g. they could both be `thingCreated`.

I think adding case-sensitive parts to URLs will add complication for both server and clients.

benfrancis:
@farshidtz wrote:
> The affordance names (thingCreation, thingUpdate, and thingDeletion) are nouns, following the TD convention for event affordances.

I have never heard of this convention. I always use nouns for property names, verbs for action names and past verb for event names. I notice that the Thing Description specification currently uses "overheated" in one place and "overheating" in another.

> I have no objection to changing those (to thingCreated, thingUpdated, and thingDeleted) if there is consensus, but that is unrelated to this PR.

OK, I've filed #258 for that.

> I think adding case-sensitive parts to URLs will add complication for both server and clients.

I don't think they necessarily have to be case sensitive, but I agree it could be a source of errors. If the affordance names at least match the wording in the URLs (#258) then I wouldn't strongly object to the URLs using snake_case while the affordance names use lowerCamelCase.

AndreaCimminoArriaga:
I think using the under score (snake_case) as separator could be a better solution since URLs are case insensitive and, therefore, using camel case may not be as suitable as the '_' separator.

farshidtz:
I'm going to change this to draft to first:
- merge TM to avoid conflicts: #253 
- bump the API version
- address #258

Issue #86, opened 2020-10-20
"Thing description" for Directory should be a "Thing Model"


mmccool:
The TD defined for a Directory service should really be a Thing Model.  We should also "template" the "base" parameter (or omit it, and say it needs to be added when the Thing Model is instantiated).

Since we are talking about a class of things, we should have a Thing Model.  However, the definition of Thing Model needs to be updated to allow specifying security scheme templates, form templates (including URL templates giving https), etc.

farshidtz:
Is that the [WoT Binding Template](https://w3c.github.io/wot-binding-templates/)?
When I search for WoT Thing Model, I only find content referring to https://www.w3.org/Submission/wot-model/

mmccool:
Thing Models used to be called "Thing Description Templates" which were basically Thing Descriptions that omitted instance-specific information.  But the TD 1.1 spec is attempting to make this a little clearer and allow URL templates (for instance) rather than just omitting hrefs.  I think all was need to do is use the following:

  "base": "https://{DIRECTORY}"

or the like.  I'd rather use a template like this rather than omitting the base so it's clear that it needs to be filled in.

egekorkan:
@farshidtz there is work going on here: https://w3c.github.io/wot-thing-description/#thing-model

@mmccool: I completely agree :)

farshidtz:
@egekorkan thanks for the link.
@mmccool it is clear now. But I think it needs more changes to become a Thing Model:

1. `base` must be replaced with placeholder, as @mmccool mentioned.

https://github.com/w3c/wot-discovery/blob/fad66ff0ff46ba72ac73420f65d8ff792103ec45/directory.td.json#L51

2. `@type` must be ThingModel
> Thing Model definitions MUST use the keyword @type at top level and a value of type string or array that equals or respectively contains ThingModel

https://github.com/w3c/wot-discovery/blob/fad66ff0ff46ba72ac73420f65d8ff792103ec45/directory.td.json#L3

3. `securityDefinitions` should be completely omitted? We don't know what schemes the instances will use.

https://github.com/w3c/wot-discovery/blob/fad66ff0ff46ba72ac73420f65d8ff792103ec45/directory.td.json#L8-L49

---

My suggestion is to keep the Thing Description as is: [An example TD](https://w3c.github.io/wot-discovery/#directory-thing-description).

relu91:
[...]

PR #253, opened 2022-01-03
Add Thing Model

farshidtz:
This PR adds a Thing Model for the directory and replaced the references to point to the API Spec section instead of the TD.

The TD instance is now changed to an example.

This is related to #86 but doesn't fully resolve it as there are still open issues:
- forms: should forms be omitted? The figure [here](https://w3c.github.io/wot-thing-description/#thing-model-concept) says "No protocol or protocol template", so the `href` part is definitely not allowed. It makes little sense to keep a form that lacks `href` since that is also the form's identifier. Maybe the figure in TD spec is out of date. I think this TM should include the forms, because in our case, the forms are normative and identical across instances.
- required interaction affordances (IA): some IAs have multiple forms, some of them optional

***
<a href="https://pr-preview.s3.amazonaws.com/farshidtz/wot-discovery/pull/253.html" title="Last updated on Jan 17, 2022, 2:56 PM UTC (cfc65fd)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-discovery/253/c46ebf7...farshidtz:cfc65fd.html" title="Last updated on Jan 17, 2022, 2:56 PM UTC (cfc65fd)">Diff</a>


egekorkan:
forms are not mandatory but can be used. Ideally, it should not be possible to make a request to a Thing just with a TM. You can template the forms or omit the base so that all URIs are relative. However, you are right, the figure at TD should not say "no protocol template"

farshidtz:
Thanks @egekorkan for the clarification. I've added the forms, without `scopes`.

Regarding the required items, how can we handle the following:
```json
"properties": {
    "things": {
        "description": "Retrieve all Thing Descriptions",
        "uriVariables": {
            "offset": {
                "title": "Number of TDs to skip before the page",
                "type": "number",
                "default": 0
            },
            "limit": {
                "title": "Number of TDs in a page",
                "type": "number"
            },
            "format": {
                "title": "Payload format",
                "type": "string",
                "enum": [
                    "array",
                    "collection"
                ],
                "default": "array"
            },
            "sort_by": {
                "title": "Comparator TD attribute for collection sorting",
                "type": "string",
                "default": "id"
            },
            "sort_order": {
                "title": "Sorting order",
                "type": "string",
                "enum": [
                    "asc",
                    "desc"
                ],
                "default": "asc"
            }
        },
        "forms": [
            {
                "href": "/things",
                "htv:methodName": "GET",
                "response": {
                    "htv:statusCodeValue": 200,
                    "contentType": "application/ld+json"
                }
            },
            {
                "href": "/things{?offset,limit,sort_by,sort_order}",
                "htv:methodName": "GET",
                "response": {
                    "htv:statusCodeValue": 200,
                    "contentType": "application/ld+json",
                    "htv:headers": [
                        {
                            "htv:fieldName": "Link"
                        }
                    ]
                },
                "additionalResponses": [
                    {
                        "description": "Invalid query arguments",
                        "contentType": "application/problem+json",
                        "htv:statusCodeValue": 400
                    }
                ]
            }
        ]
    }
},
```
There are two forms for the `things` property. Only the first one is mandatory.

I can think of the following:
```json
"tm:required": [
        "#/properties/things/forms/0"
],
```

(This is a very complicated way of defining what is required. Why not simply allow setting `"tm:required": true` inside the definition itself, like in OpenAPI?)

egekorkan:
Your solution is indeed correct but results in this complicated notation. Your alternative solution can be also possible in the future. I think that the reasoning behind the current solution was to allow something like `minimum` of a DataSchema to be required, which is not an object and cannot contain `tm:required : true`

farshidtz:
From the Discovery meeting: Give more time to review.

farshidtz:
The following assertion in TD spec is too strict. THe derived TD can use a type which is of Thing subtype.
> The tm:ThingModel value of the top-level @type MUST to be replaced by the value Thing in the Partial TD instance.
> https://w3c.github.io/wot-thing-description/#thing-model-td-generation

egekorkan:
@farshidtz could you explain what you mean by:

> THe derived TD can use a type which is of Thing subtype.

farshidtz:
> @farshidtz could you explain what you mean by:
> 
> > THe derived TD can use a type which is of Thing subtype.

@AndreaCimminoArriaga can explain better as it was suggested by him.

Our concern is that the assertion requires replacing `tm:ThingModel` type to `Thing` and nothing else. 

- What if the type is meant to be a `discovery:ThingDirectory` (see the [ontology](https://w3c.github.io/wot-discovery/#exploration-mech)) or [`saref:TemperatureSensor`](https://w3c.github.io/wot-thing-description/#saref-state-annotation-example)? Those custom types can be subclasses of a `Thing` as defined in their ontologies without the need to explicitly have a `Thing` type set in the TD itself.
- What if type is an array which may already include the instance type? The type of TM being added by this PR:
```json
    "@type": [
        "tm:ThingModel",
        "ThingDirectory"
    ],
```

egekorkan:
I understand now. I think the assertion in the TD spec should be changed to simply removing ` "tm:ThingModel"` from the value(s) of `@type`. Would this be acceptable?

farshidtz:
> I think the assertion in the TD spec should be changed to simply removing ` "tm:ThingModel"` from the value(s) of `@type`. Would this be acceptable?

Yes, I think that would work.

AndreaCimminoArriaga:
Sorry for the late response, I think @farshidtz explained it perfectly and the solution proposed by @egekorkan will do the work

farshidtz:
I've flattened the API spec chapter to only include the TM. I removed the example TD because it was making the spec very long and it was very hard to maintain and keep it consistent with the TM. 

Future forward-incompatible modifications to the TM should increment the version number to ensure validity of example TD. 

The TD has been moved to `implementation/example-directory-oauth2.td.json`

Issue #245, opened 2021-12-06
Disallow SPARQL UPDATE operations


farshidtz:
The spec indicates that `UPDATE` is out of the scope of the SPARQL Seach API: https://github.com/w3c/wot-discovery/blob/40267a23442a44beca4724f4d7495eeba2e9deed/index.html#L1834

In the discovery call today, it was proposed that this statement be changed to explicitly disallow this operation. An appropriate status code should be set in response to such query, perhaps 403 Forbidden.

AndreaCimminoArriaga:
I agree, the UPDATE query would allow registering TDs (or modifying existing ones) bypassing the endpoints for such purpose.

mmccool:
The SPARQL system has different endpoints for UPDATE, so we could simply not support the UPDATE endpoint.  @AndreaCimminoArriaga will investigate.

wiresio:
In Apache Jena an update on the /query endpoint results in http 400 (bad request)

AndreaCimminoArriaga:
I have created a PR for this issue, but I have a parallel question. Shall we add examples of JsonPath, Xpath and SPARQL queries on the spec?

farshidtz:
I propose we close this issue since the PR has been merged.

> I have created a PR for this issue, but I have a parallel question. Shall we add examples of JsonPath, Xpath and SPARQL queries on the spec?

We have separate issues for this:
- #81 
- #141

PR #259, opened 2022-01-17
Explicitly disallow SPARQL Update queries

AndreaCimminoArriaga:
Related to issue #245

### DIFF
--- a/index.html
+++ b/index.html
@@ -1845,13 +1845,14 @@ img.wot-diagram {
                             The support for SPARQL search using HTTP <code>POST</code> method
                             at `/search/sparql` endpoint is OPTIONAL.
                         </span>
-                        <code>UPDATE</code> queries are out of the scope for the API.
+			SPARQL <code>UPDATE</code> queries are disallowed.
                         <span class="rfc2119-assertion" id="tdd-search-sparql-resp">
                             A successful response MUST have 200 (OK) status, and depending on the type
                             of query contain by default as Content-Type header <code>application/ld+json</code>
                             for <code>CONSTRUCT</code> and <code>DESCRIBE</code> queries or
                             <code>application/json</code> for <code>SELECT</code> or <code>ASK</code>.
                             The response body MAY contain TD fragments or a set of TDs depending on the query.
+			    A request with a query different from <code>SELECT</code>, <code>ASK</code>, <code>CONSTRUCT</code> or <code>DESCRIBE</code> MUST return a response 501 (Not Implemented).
                         </span>
                         The semantic search with SPARQL is specified as `searchSPARQL` action in
                         [[[#directory-thing-description]]].


index.html
```diff
@@ -1845,13 +1845,14 @@ <h4>Semantic search: SPARQL</h4>
...
-                        <code>UPDATE</code> queries are out of the scope for the API.
+			SPARQL <code>UPDATE</code> queries are disallowed.
                         <span class="rfc2119-assertion" id="tdd-search-sparql-resp">
                             A successful response MUST have 200 (OK) status, and depending on the type
                             of query contain by default as Content-Type header <code>application/ld+json</code>
                             for <code>CONSTRUCT</code> and <code>DESCRIBE</code> queries or
                             <code>application/json</code> for <code>SELECT</code> or <code>ASK</code>.
                             The response body MAY contain TD fragments or a set of TDs depending on the query.
+			    A request with a query different from <code>SELECT</code>, <code>ASK</code>, <code>CONSTRUCT</code> or <code>DESCRIBE</code> MUST return a response 501 (Not Implemented).
                         </span>
```
farshidtz, line 1856: There are two MUST and one MAY assertion keywords in a single assertion block. Not sure if this is strictly disallowed, but it will make testing of implementations harder because there is only one ID for the three assertions.

farshidtz: > "return a response 501 (Not Implemented)"

Is this coming from the SPARQL spec?
Status [501](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/501) kind of implies that the server is not implementing it right now, and may do it in future.

farshidtz (commented):
"return a response 501 (Not Implemented)"

index.html
```diff
@@ -1845,13 +1845,14 @@ <h4>Semantic search: SPARQL</h4>
                             The support for SPARQL search using HTTP <code>POST</code> method
                             at `/search/sparql` endpoint is OPTIONAL.
                         </span>
-                        <code>UPDATE</code> queries are out of the scope for the API.
+			SPARQL <code>UPDATE</code> queries are disallowed.
```
mmccool, line 1848: Probably should be an assertion, and should be a statement about the content of the query itself.  Maybe:
"The SPARQL query MUST NOT include <code>UPDATE</code> statements."  Or it can be dealt with in a different way below (by stating what IS allowed, "a request with a query different".).  We could also add "In particular, <code>UPDATE</code> is disallowed."

index.html
mmccool: Agree, these need to be broken into multiple spans, one per assertion, with unique ids.  The tooling looks for a particular keywords in each assertion so if there is more than one it will complain/be unable to track tests.

AndreaCimminoArriaga: Regarding splitting the assertion I agree as well, I will put myself on splitting the assertions. 

Regarding the code, SPARQL says [this](https://www.w3.org/TR/sparql11-protocol/#query-failure) about failure response codes. Taking  this sentence `A protocol service may use other 4XX or 5XX HTTP response codes for other failure conditions, as per HTTP.` into account and considering the fact that in other functionalities we used 501 for indicating not implemented I though on using the same for the UPDATE queries. However I have no problems in using a different code (e.g., 500)

farshidtz: In other cases, we suggest the use of 501 when a server doesn't implement an expected feature.

I think we should classify it as a [4xx](https://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html#sec10.4) error, because it is cause by a bad user request and not an error from the server typically classified as [5xx](https://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html#sec10.5). The server recognizes the request, **is capable of serving it**, but refuses to do so. Please correct me if I'm wrong.

It could be [403](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/403):
> The HTTP 403 Forbidden response status code indicates that the server understands the request but refuses to authorize it.
>
> This status is similar to 401, but for the 403 Forbidden status code re-authenticating makes no difference. The access is permanently forbidden and tied to the application logic, such as insufficient rights to a resource.

Looking at the SPARQL spec, [500](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/500) would at first seem like an appropriate choice, but I believe "refusal to execute a query" is caused by an internal server error. 
> The HTTP response codes applicable to an unsuccessful query operation include:
>
> - 400 if the SPARQL query supplied in the request is not a legal sequence of characters in the language defined by the SPARQL grammar; or,
> - 500 if the service fails to execute the query. **SPARQL Protocol services may also return a 500 response code if they refuse to execute a query.** This response does not indicate whether the server may or may not process a subsequent, identical request or requests.
>
> The response body of a failed query request is implementation defined. Implementations may use HTTP content negotiation to provide human-readable or machine-processable (or both) information about the failed query request.
>
> A protocol service may use other 4XX or 5XX HTTP response codes for other failure conditions, as per HTTP.

index.html
```diff
@@ -1845,14 +1845,15 @@ <h4>Semantic search: SPARQL</h4>
...
                         </span>
-                        <code>UPDATE</code> queries are out of the scope for the API.
-                        <span class="rfc2119-assertion" id="tdd-search-sparql-resp">
-                            A successful response MUST have 200 (OK) status, and depending on the type
-                            of query contain by default as Content-Type header <code>application/ld+json</code>
-                            for <code>CONSTRUCT</code> and <code>DESCRIBE</code> queries or
-                            <code>application/json</code> for <code>SELECT</code> or <code>ASK</code>.
-                            The response body MAY contain TD fragments or a set of TDs depending on the query.
+                        <span class="rfc2119-assertion" id="tdd-search-sparql-resp-select-ask">
+                            A request with a query <code>SELECT</code> or <code>ASK</code> MUST return a response 200 (OK) status, and contain by default as Content-Type header <code>application/json</code>.
```
mmccool, line 1849: request -> successful request

index.html
```diff
@@ -1845,14 +1845,15 @@ <h4>Semantic search: SPARQL</h4>
...
                         </span>
+			 <span class="rfc2119-assertion" id="tdd-search-sparql-resp-describe-construct">
+                            A request with a query <code>CONSTRUCT</code> and <code>DESCRIBE</code> MUST return a response 200 (OK) status, and contain by default as Content-Type header <code>application/ld+json</code>.
```
mmccool, line 1852: request -> successful request

index.html
```diff
@@ -1845,14 +1845,15 @@ <h4>Semantic search: SPARQL</h4>
...
+			 </span>
+			<span class="rfc2119-assertion" id="tdd-search-sparql-error">
+			     A request with any query different from <code>SELECT</code>, <code>ASK</code>, <code>CONSTRUCT</code> or <code>DESCRIBE</code> MUST return a response 501 (Not Implemented).
```
mmccool, line 1855: what about query with a syntax error that has an UPDATE?

mmccool:
Discussion: on last point, 400 Bad Request could also be used, and would also cover syntax errors.  403 implies that proper credentials have not been provided, which is not the case.  400 is already used for syntax errors in the spec, which makes it irrelevant whether we check syntax or for UPDATE first.

PR #261, opened 2022-01-19
Update discovery-context.jsonld

AndreaCimminoArriaga:
Modified the context so it works again, some keywords must be discussed

### DIFF
--- a/context/discovery-context.jsonld
+++ b/context/discovery-context.jsonld
@@ -13,13 +13,15 @@
       "@id": "discovery:ThingCollection",
       "@type": "@id"
     },
-	"total": "discovery:contains",
-	"members": [
-	  "td:Thing"
-	],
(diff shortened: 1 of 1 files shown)


mmccool:
So if I understand this correctly, this updates "next" so it is typed as a URI.  Were there any other keyword changes?  Can you summarize what else needs to be discussed?

AndreaCimminoArriaga:
This update fixes the keys `next` and `total`. However, other two keys generate conflicts in the context that are the `@id` and `member`, for the moment I have erased them so the context has no errors and can be used for the round-tripping. I'm not sure if we need to discuss about them (we already agreed on having this alternative way of paginating), they just need to be fixed according to the needs of @wiresio who proposed them and upload the ontology with them.

On the other hand, I would like to discuss today some other terms from our ontology like create and modified since I saw that the new TD spec already includes this terms.

Issue #208, opened 2021-06-30
Read-Only Directories


RoboPhred:
I am building a device hub around TD.  Currently, it communicates available TDs by simply sending an array of all known TDs to a predetermined url known by my specific client app, but I am interested in switching to the Discovery spec for greater compatibility.

However, looking through the spec, it seems the create / update / delete operations are not optional.  Since my hub represents a source of real devices (zwave in this case), creation or modification of their spec is undesirable.  The closest I could get is setting the hub into inclusion mode for a new zwave device, but the device capabilities (and the resulting TD) is not left up to the user.

After reviewing the options in the Discovery spec, I don't seem to have a clear path forward:
- A Directory (as I currently understand it) will not work, as I do not want to provide the capability of adding arbitrary TD or allowing for arbitrary updates on existing things.
- A well known url will not work, as I can only have one, and it can only return a single TD.  I have multiple devices I need to return.
- mdns will not work as I cannot broadcast a url path alongside it, and all my things are under the hub's domain / server.

I suppose I could implement the CRUD actions and return FORBIDDEN for all of them regardless of operation, but that seems a bit cheaty...

My use case would be much better served if there was a way to advertise a single thing, and have that thing return additional TDs, just like a directory but without the contractual mutation support.

relu91:
I would support lessening a little bit the spec to allows TDD to be just a queryable collection of TDs. It makes the mandatory implementation requirements smaller hence simpler to realize. Besides, I know that we are not aiming for cross-protocol compatibility, but a smaller set of functionalities is also easier to port. 

An additional point would be that the webthings.io gateway would be easier to align. Maybe @benfrancis can expand on this.

benfrancis:
Yes we have the same situation with WebThings Gateway.
1. Thing Descriptions are generated by the gateway to represent a real underlying device which may use another protocol like Zigbee or Z-Wave, it's not possible for a Consumer to just add an arbitrary Thing Description to the gateway's directory
2. It doesn't make sense for Consumers to be able to edit a Thing Description because they could for example add a property which doesn't exist in the underlying device, leaving the Web Thing in a broken state

+1 for allowing read-only directories.

relu91:
Going back to this issue. In this PF (among the other projects), I was thinking to add a TDD interface to node-wot. But again the mandatory updates methods make no sense in that context, cause the WebThings can be only registered by node-wot itself. 

Also, why can a TDD  only be a collection of TDs? Why is JSONPath mandatory? 

[...]

PR #267, opened 2022-01-29
Disambiguate assertions in registration API

### LINKED ISSUE #266: Minimum requirements for a TDD
There are some issues about the minimum requirements on search but I could not find one about requirements in general for any TDD. If someone gives me the URL of a TDD, am I guaranteed to get an answer to an HTTP GET request to `/things`? How about the other operations/methods and endpoints etc.? I have mentioned a bit of this at https://github.com/w3c/wot-discovery/issues/178.

It would be nice to have `tm:required` on mandatory affordances of the TDD TM if this is indeed the case. Also, from the text it seems that I do not have to implement retrieval but I have to implement update and registration. To be clear, here is what retrieval has as assertion:
> A TD MUST be retrieved from the directory using an HTTP GET request at /things/{id} endpoint, where id is the unique TD identifier. A successful response MUST have 200 (OK) status, contain application/td+json Content-Type header, and the requested TD in body. 

which does not mean that retrieval is mandatory but how it should be implemented is assertive. Whereas the creation says>
> The API MUST allow registration of a TD object passed as request body.

which is also not very clear but I can understand it as creation being mandatory.

There is also the following assertion:

> The Registration API MUST provide create, retrieve, update, delete, and listing (CRUDL) interfaces. 

However, it is not watertight that these interfaces are for TDs (a bit of a weird interpretation but possible).

In any case, is it allowed to have a TDD that is only for retrieval and another one that is only for creation and updates. Currently, the only way to have such levels is having different security levels.

farshidtz:
Closes #266

***
<a href="https://pr-preview.s3.amazonaws.com/farshidtz/wot-discovery/pull/267.html" title="Last updated on Jan 31, 2022, 3:16 PM UTC (ea924db)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-discovery/267/567c286...farshidtz:ea924db.html" title="Last updated on Jan 31, 2022, 3:16 PM UTC (ea924db)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -976,7 +976,7 @@ img.wot-diagram {
                         <h4>Retrieval</h4>
                         <p>
                             <span class="rfc2119-assertion" id="tdd-reg-retrieve">
-                                A TD MUST be retrieved from the directory using an HTTP `GET` request
+                                The API MUST allow retrieval of an existing TD using an HTTP `GET` request
                                 at `/things/{id}` endpoint, where `id` is the unique TD identifier.
                             </span>
                             <span class="rfc2119-assertion" id="tdd-reg-retrieve-resp">
--- a/index.html
+++ b/index.html
@@ -880,7 +880,8 @@ img.wot-diagram {
                     </p>
                     <p>
                         <span class="rfc2119-assertion" id="tdd-reg-crudl">
-                            The Registration API MUST provide create, retrieve, update, delete, and listing (CRUDL) interfaces. 
+                            The Registration API MUST provide interfaces to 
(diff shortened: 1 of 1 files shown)


egekorkan (approved):
Thanks 👍

benfrancis:
"The Registration API MUST provide interfaces to create, retrieve, update, delete, and list (CRUDL) TDs."

@farshidtz Does this mean that read-only (#208) directories are not allowed?

benfrancis:
@farshidtz OK, if in WoT Discovery 1.0 all directories have to be writable, then that's going to be a problem for WebThings (and I suspect most IoT gateways). I have followed up in #208.

Issue #263, opened 2022-01-20
Profile needs a simple self descriptive mechanism without huge implementation demands


mlagally:
Does a simple thing have to implement a directory service / implement a discovery API?

benfrancis:
My proposal in https://github.com/w3c/wot-profile/issues/112 for which I have a PR at https://github.com/w3c/wot-profile/pull/121 is that all Things conforming to the Core (HTTP+JSON) Profile must as a minimum have their Thing Description retrievable from an HTTP URL. This would allow for [direct discovery](https://w3c.github.io/wot-discovery/#introduction-direct) (e.g. copy & pasting a URL or using a QR code or Bluetooth beacon).

There was also a suggestion that the Core Profile should make the [Well-Known URI introduction mechanism](https://w3c.github.io/wot-discovery/#introduction-well-known) mandatory, but there are issues with that, discussed in https://github.com/w3c/wot-profile/pull/121.

There was also a suggestion that the Core Profile should make support for the [Directory Service API](https://w3c.github.io/wot-discovery/#exploration-directory-api) mandatory for conformant Consumers, but that would significantly increase implementation complexity.

mmccool:
- "Peer to Peer" discovery is currently defined here: https://w3c.github.io/wot-discovery/#exploration-self
- This just states that a GET retrieves a TD from a URL directly.  It needs to be paired with an "Introduction" mechanism to determine a URL
- The simplest Intro mechanism to pair with this is the ["well-known" URI](https://w3c.github.io/wot-discovery/#introduction-well-known)
- So, I would suggest that Profiles require both the well-known URI introduction mechanism and the self-description mechanism described above.
- HOWEVER this does not currently handle the use case of multiple Things on one server.  This could happen if there are multiple services running on a hub, for instance.  These cases should use a directory, but since queries are not now required for directories, this is not a "huge" implementation burden.  The question is if we need something simpler for small devices with a small number of TDs.  One use case that comes to mind is a dual-outlet plug with separate Things for each outlet.  Doing this makes it easier to set up automations that expect to talk to a single on-off device.  I was thinking that we should just extend self-description (section 6.1) to allow it to return a JSON Array of TDs.  It would be easiest if it *always* returned an array, in my opinion... will discuss today.   This would only be recommended for a small number of TDs, up to 3 or 4.  Beyond that a directory should be used, as it supports pagination, etc.
[...]

Issue #249, opened 2021-12-13
Multiple Responses from Peer-to-Peer Discovery


mmccool:
It was noted that when "guessing" the URL to access TDs from self-describing Things given just the IP address that we have to deal with the case of endpoints that host *multiple* Things, like a gateway (which could have several services acting as Things, including proxies) or even endpoint with multiple sub-devices (e.g. a geolocator, a temp sensor, a humidity sensor, etc all in one device).

Proposed: the return value at .well-known can be an array of TDs.  If this proposal makes sense, we should also discuss whether it should always be an array (e.g. should square brackets always be used even if there is just one element) or if the array notation is only necessary if there are multiple TDs.  We could also discuss making this consistent with the TDD listing API, whether we need pagination (might be useful for gateways with a bunch of services, including, for instance, proxies for all devices on the LAN), etc.  This aligns with my initial thought of making .well-known a special case of a TDD.

Assumption: authentication determines the identity of the requestor and whether they have access to any of the things or a subset.  Only the subset they have the right to view should be returned.  If the array would have been empty (no rights to view anything) then an Unauthorized error should be returned rather than an empty array.

mmccool:
Discussion:
- Could also return one TD that links to other TDs.  This gives a TD similar functionality of CoRE-RD, but without bringing in another spec (and format that needs to be parsed by a consumer).  Would be helpful to identify with a special "ThingCollection" `@type`.  Or we can use the existing ["ThingLink" type](https://w3c.github.io/wot-discovery/#exploration-mech)
- If we return an array... in DNS service type, is it a Thing or a Directory?
- Maybe we should use CoRE-RD format for .well-known?
- We have an IANA content type for TDs, but not for "collections of TDs".  So what would the content type of the .well-known endpoint be?
- in Node-wot script, can have expose multiple Things from the same script, servient exposes itself using a TD.  This leads to different cases for one and multiple TDs.  Do we switch between them?  Or just always use ThinkLinks?  Disadvantage: need to fetch two TDs, so two network transactions.  If we return the TD itself for "singleton" Things then we avoid this problem for the simple case (one Thing for a given IP).

mmccool:
I'm leaning towards:
- Using a ThinkLink TD, but only when there are multiple Things at a given IP address.

benfrancis:
I was under the impression that this use case was already solved by the text in [section 5.2](https://w3c.github.io/wot-discovery/#introduction-well-known):

[...]

PR #269, opened 2022-01-31
Permit use of TD Links in well-known URIs

mmccool:
We have discussed using TD Links in TDs returned by well-known URIs to support self-description by endpoints hosting multiple TDs.  This PR adds the following:
- An informative comment in the well-known URI section discussing this use case

To do:
- An informative comment in section 6.1

To discuss:
- Whether we need restrictions/types on TD Links, e.g. they may not point at TDDs, but only TDs, in order to address Issue https://github.com/w3c/wot-discovery/issues/263

### DIFF
--- a/index.html
+++ b/index.html
@@ -306,6 +306,13 @@ img.wot-diagram {
                     When a request is made at the above Well-Known URI, the server MUST return
                     a Thing Description as prescribed in [[[#exploration-self]]].
                 </span>
+		It is sometimes the case that a well-known URI
+		needs to index multiple TDs.  This can be the case,
+		for example, when a URI is inferred from
+		an IP address but multiple Things are hosted
+		at that IP address.  In this case, a single
+		TD is still returned by the process but that
+		TD can use TD Links to reference other TDs.
             </p>
             <p class="ednote" title="Registration of Well-known URI">
                 The service name in Well-Known URI (<code>wot-thing-description</code>) is tentative.


mmccool:
The current definition of "ThingLink" applies to the entire TD, but only has a "describedby" link type which is not really want we want.  I would like to propose adding "thing" and "directory" types to links (to match the ones in CoreRD) so that a TD can be used, in essence, like an additional introduction mechanism.

mmccool:
Terminology updated to reference "Thing Links" (which are a type of TD) but we may want to have other link types than just "describedby".

mmccool:
Figure 4 may also have to be updated to allow Self-describing Thing to return a Thing Link.

farshidtz:
IMO, the link relation should specify the relation. There is already a `type` field which specifies the media type `application/td+json` and can be extended with parameters if needed e.g. `application/td+json;type=directory`

For the DNS-SD service type, we default to Thing and allow setting directory:
> `type`
Type of the Thing Description, i.e. Thing or Directory. If omitted, the type is assumed to be Thing.


> Figure 4 may also have to be updated to allow Self-describing Thing to return a Thing Link.

It already does, because Thing Link is a subclass of Thing (see figure 5)

mmccool:
I actually deleted my comment about Figure 4 before your comment as I noticed the subclass :).  Anyway, let's discuss today.

mmccool:
The relation type "item" is already mentioned in https://w3c.github.io/wot-thing-description/#link, and means "a TD that is a member of a collection" which is appropriate.

mmccool:
See also discussion in scripting on how to deal with [composite TDs](https://github.com/w3c/wot-scripting-api/issues/363)

mmccool:
Points to decide:
- Expand Thing Link or add a new "type" of TDs.  Resolution: expand Thing Link
- Do we use a different relation type than "describedby", and if so, what type?  The "item" implies a collection but there may not be a collection per se (it may only be a set of services that happen to be hosted on the same device but are otherwise unrelated).  The relation "hosts" is used in Core RD but is sort of generic.  The type "thing" is used for Core RD to point at things (and there is also "directories", but "describedby" does cover that for directories referring to others).  Resolution: use either "hosts" or "external" for this relation.   The first should be used when the target TD is under the same context IRI, and "external" when it is hosted elsewhere.  Note: does not replace "describedby".  This is a different use case.  Note: both hosts and item assume the same "context IRI", so either is only appropriate for self-description where the TDs are, in fact, hosted on the same device.  If you have to go somewhere else can use a "describedby" or better, an "external" link, however.  We can allow "hosts" to be relative to the current TD, whereas "external" should be absolute.
- Do we depend on the content type alone to resolve relation types?  Alone: no.  In addition: redundant.  Note: relation type is optional.  Resolution: use content type "application/td+json" only.
- How do we handle Thing Link in discovery process.  Resolution: follow links (and expand collections), and add them to the "bag" of TDs discovered; "collection" relationships are still described in the TDs themselves, so no information is lost)

mmccool:
I have expanded the definition of Thing Link to allow "hosts" and "external" relation types.  This required a number of new assertions and a refactoring of existing ones.  I also added an example and a use case distinguished from "describedby".  Still to do: update the intro under "Architecture" and Figure 1 to show Thing Links and the "expansion" of Links.  This section needed to be filled in anyway...

index.html
```diff
@@ -567,6 +646,22 @@ <h1>Exploration Mechanisms</h1>
...
+		as the permitted single TD, but can then reference multiple TDs.
+		</li>
+	        </ul>
+		Note that the `hosts` and `external` relation type, and Thing Links in general, do not
+		necessarily describe collections of TDs.  Things may be co-hosted under a common base URL
+		and/or physical device (with a single IP address, for example) 
+		without being directly related to each other.  
+		If a set of Things <em>are</em> part of a collection, additional Links should be provided
+		in the individual TDs to describe such relations, for example using `collection` and
+		`item` relations.
```
farshidtz, line 663: I think this paragraph will add a lot of confusion because it directly overlaps with the functionalities of directories and other relation types defined above it. In this spec, we have numerous references to "collection", a JSON-LD class called "ThingCollection" and listing format "collection"; all referring to what is stored in the directory or the list of TDs.

Given that, can a Link with "collection" rel point to a directory? Because TD [spec](https://w3c.github.io/wot-thing-description/#link) says: "collection: Points to a collections of Things". What will be the difference from a TD with type ThingDirectory and "describedby" link referring to that directory?

Looking at the RFC that defined that relation:
>   When included in a resource that represents a member of a collection,
   the 'collection' link relation identifies a target resource that
   represents a collection of which the context resource is a member.
https://www.rfc-editor.org/rfc/rfc6573.html#section-2.2

If I understand correctly, this means that the TD pointing to that link must be part of the linked "collection". 

> When included in a resource that represents a collection, the 'item'
   link relation identifies a target resource that represents a member
   of that collection.
https://www.rfc-editor.org/rfc/rfc6573.html#section-2.1

Is the linked "item" a member of that parent TD "collection"? Is the parent TD even a collection?

farshidtz (changes requested):
Thanks @mmccool. I think this is very clear. I just have some concerns about the paragraph about item and collection; please see the inline comment.

index.html
mmccool: ok, I'll think about it.  I could just take this out but I do want to emphasize that is this just an example.  It's a lot easier to do the iterative discovery process if there is no important information stored in Thing Links, so they can be elided as part of the process.  BTW I just expanded the arch intro to describe that process as well (including updating the diagram).

relu91:
Thank you for bringing this to the table and sorry for being absent in the latest Discovery calls. Some points: 

Introducing the ability to point to a set of links in the ThingLink may solve the [node-wot use case](https://github.com/w3c/wot-discovery/issues/208#issuecomment-932728032) for read-only Thing Description Directory. So to expand a little bit the use case, node-wot is a servient runtime that is self-described by a Thing Description. Since the runtime may `host` one or more WebThings, the idea was to use the servient TD as an entry point to discover the list of WebThings hosted there. Now, reading the current proposal, I'm not sure if I have been allowed to mix Thing and ThingLink types in a TD. To be concrete here is the TD that I would like to host in node-wot:
```json
{
   "title": "servient", 
   "id": "urn:dev:12321"
   "@type": ["Thing", "ThingLink"], 
   "links": [ {  
                    "rel": "hosts",
                    "href": "thing1",
                    "type": "application/td+json"
                 },
                 {
                   "rel": "hosts",
                   "href": "thing2",
                   "type": "application/td+json"
                 }],
   "actions" : {
      "shutdown": {},
      "setLogLevel": {},
      "runScript": {},
   }
}
```
**Note**: the assumption is that the servient TD is discovered using another introduction mechanism (e.g., DNS-SD). 

Does this fulfill the requirements in this PR? 

Another point: As you probably already know, with the newly introduced [Thing Model Composition](https://w3c.github.io/wot-thing-description/#thing-model-composition) TDs may refer to their subparts using links with rel type `item`. I would harmonize that solution with what we are defining in this PR, which means that we should stick with the same rel type in both specs. Personally, I'm ok to use `host` instead of `item` so that we could discover TD subparts following the directions and the assertions given in the Discovery spec.

benfrancis:
Does this just re-invent directories..?

@relu91 wrote:
> To be concrete here is the TD that I would like to host in node-wot:
> 
> ```json
> {
>    "title": "servient", 
>    "id": "urn:dev:12321"
>    "@type": ["Thing", "ThingLink"], 
>    "links": [ {  
>                     "rel": "hosts",
>                     "href": "thing1",
>                     "type": "application/td+json"
>                  },
>                  {
>                    "rel": "hosts",
>                    "href": "thing2",
>                    "type": "application/td+json"
>                  }],
>    "actions" : {
>       "shutdown": {},
>       "setLogLevel": {},
>       "runScript": {},
>    }
> }
> ```

I really like this design for describing a collection of Things, and you will recall that it was one of the proposals for describing a Directory. But the conclusion at the time was that a Thing Description should be considered to be fairly static resource, so providing a dynamic list of links inside a Thing Description wasn't a good idea.

The current discovery process is farily simple
1. The Introduction mechanism results in a URL
2. That URL can either resolve to a Thing Description (describing a single Thing) or a Directory Description (providing access to a collection of Things)

As far as I can tell this PR adds an additional mechanism for describing a collection of devices in addition to a Directory. Now that URL from the introduction mechanism could resolve to:
1. A Thing Description
2. A Directory Description
3. A ThingLink linking to a single Thing
4. A ThingLink linking to multiple Things

This adds quite a lot of implementation complexity for Consumers and doesn't actually solve the original problem which I think may have triggered this issue which related to the well-known URI introduction mechanism.

The problem (identified in https://github.com/w3c/wot-profile/pull/121#issuecomment-970404999) is simply that if the well-known URI mechanism is made mandatory for the Core Profile then that has the side-effect of also making support for the Directory Service API mandatory for conformant Consumers, since if a well-known URI may resolve to a Directory Description then a Consumer will need to know what to do with it.

Extending the well-known URI introduction mechanism so that it may also resolve to a ThingLink linking to a single Thing or a ThingLink linking to multiple Things doesn't solve that problem, it just adds further implementation complexity.

Two ways of solving that problem are:
1. Simply not making the well-known URI introduction mechanism mandatory in the Core Profile (which is my recommendation)
2. Making support for the Directory Service API mandatory in the Core Profile (which I'm also fine with, but which adds implementation complexity for Consumers, so it might make sense for it to be part of a separate profile instead)

I may have misunderstood and you're actually trying to solve another problem, but I think this PR just adds a duplicate mechanism for describing collections of Things.

The issue of read-only directories should be probably be treated separately in #208 and is really just a case of making write operations optional in the spec.

relu91:
> If we use collection/item here I would be removing the "collection" master TD which is probably not desirable.

But in principle, you can still recover it using the backlink contained in the collection items, right? Also, why remove the master collection TD? 

> I would prefer that Thing Links include NO affordances. That makes it a lot easier to reason about them and replace them with their referents during the Discovery process (now described under Architecture as part of this PR). If they have affordances, replacing them with their referents removes those affordances, which does not really fit that third use case. If we use "hosts" it can Link to both the "master" (collection) TD /and/ each "item" TD so the collection structure is maintained in the result of the Discovery.

It is a little bit annoying to be presented with just an empty TD that refers to the "master" collection and its items (you need an additional fetch), but I understand your previous concerns. 

> Does this just re-invent directories..?

TBH, it kinda is. Indeed, in my mind, it covers the same use case (node-wot needs to present its WebThings). However, I know that the intention here is to have a simplified solution for small devices that hosts multiple WebThings. Therefore, in such use cases, the number of hosted WebThings will be quasi-static and they can be presented as links in a static TD document. 

> The problem (identified in w3c/wot-profile#121 (comment)) is simply that if the well-known URI mechanism is made mandatory for the Core Profile then that has the side-effect of also making support for the Directory Service API mandatory for conformant Consumers, since if a well-known URI may resolve to a Directory Description then a Consumer will need to know what to do with it.

I actually missed that discussion, but as I said above I think we have other valid use-cases. 

> The issue of read-only directories should be probably be treated separately in #208 and is really just a case of making write operations optional in the spec.

Ok, point taken. I just want to mention that if read-only TDDs will be implemented then the node-wot use case is covered. Basically, the runtime will expose the following TD:
```json
{
   "title": "servient", 
   "id": "urn:dev:12321"
   "@type": ["Thing", "ThingDirectory"], 
   "properties": {
      "things": {},
    }
   "actions" : {
      "shutdown": {},
      "setLogLevel": {},
      "runScript": {},
   }
}
```

benfrancis:
@relu91 wrote:
> I just want to mention that if read-only TDDs will be implemented then the node-wot use case is covered. Basically, the runtime will expose the following TD:
> 
> ```json
> {
>    "title": "servient", 
>    "id": "urn:dev:12321"
>    "@type": ["Thing", "ThingDirectory"], 
>    "properties": {
>       "things": {},
>     }
>    "actions" : {
>       "shutdown": {},
>       "setLogLevel": {},
>       "runScript": {},
>    }
> }
> ```

This is exactly what we need for WebThings Gateway too. A Thing Description which can describe the physical gateway itself as a Thing (with actions like "reboot"), but also the list of Things it hosts (via the things property). I would have liked that to be a list of links rather than a property, but that's not the solution that was landed on for directories.

I do agree that a directory is overkill for the "multiple outlets on a power strip" use case, but I'm not sure they're actually multiple Things. I would personally model that as one Thing, but it could also be modelled as a top level Thing (with its own Thing Description) which links to other sub-Things.

mmccool:
General comments, we can discuss more in the upcoming profile mtg:
- the Think Link approach is specifically for the use case of one small device hosting a small number of *static* Things, and for when you want to support self-discovery.  If you want to support a dynamic set of Things, then you really need an API to manage the set, notifications of updates, etc; that's the use case for the TDD service.   Please don't use the Thing Link approach for dynamic resources, that's not its use case (I think I said that somewhere in the text, also...).
- I think it would be fine in Profiles to disallow directories in some cases.  However, the architecture means that the Thing can decide whether it wants to host a TDD service or not.  It's not forced to; if it is, please point out specifically what we would have to change in the spec. 
- I bring up the power bar example because I ran into this aggravating issue when trying to deal with automations in Home Assistant and Smart Things.  Basically lots of automation rules assume that they are dealing with "singular" items, i.e. one on-off switch, one sensor value.  They get confused by devices that have multiple actuators or sensor values.  So rather than have one big Thing with a lot of affordances in this case that you then have to index into, it's better to have a bunch of simpler Things that you can reference directly imo.  Simply less chance of other parts of the system getting confused.  As noted, "collection/item" can also do this.  However using "hosts" means that the discovery process itself can "flatten" the set so the automation system would just see a bunch of devices.
- Which gets to the last point, mixing in affordances into the Thing Link.  I would prefer not to do this as I'd like the option to allow the Discovery process to "expand" Think Links by replacing them with their referents.  If you want to provide "host control" affordances then create one and add it to the set of links.

index.html
```diff
@@ -578,12 +711,26 @@ <h2>Self-description</h2>
...
+	    dynamically over time, 
+	    a <a>Thing Description Directory</a> (TDD)
+	    can be used instead; in such a case the TD returned is that 
+	    of the TDD.
+	    If a small fixed number of TDs need to be returned,
+	    particularly on small devices that do not want to host the
+	    full TDD interface, a single TD can be returned, marked as 
+	    a Thing Link, which link to multiple TDs hosted  
+	    at different
+	    URLs.
```
farshidtz, line 732: This gives the false impression that Thing Link and TDD are alternative ways of returning multiple TDs. But in fact, Thing Link serves the very specific purpose of allowing a single TD to **reference** one or more.

images/overview.drawio
```diff
@@ -1 +1,191 @@
...
+              <mxPoint x="700" y="458" />
+              <mxPoint x="700" y="540" />
+              <mxPoint x="465" y="540" />
+            </Array>
+          </mxGeometry>
+        </mxCell>
+      </root>
+    </mxGraphModel>
+  </diagram>
+</mxfile>
```
farshidtz, line 191: This figure is confusing:
1. The diagram makes it look like Thing Link and TD are different entities. Thing Link is a type of TD. 
2. Exploration enclosure is covering only directory. Instead it should cover everything after URL is found. Note: "These URLs do not themselves contain metadata, but are used in the second stage, "Exploration", to actually fetch metadata."

I'd instead point from URL to TD (self-described Thing) and use empty arrow heads to divide it using from there into Thing Directory (pointing back to TD) and Thing Link (pointing back to URL).

benfrancis:
Thanks for these clarifications @mmccool. I doubt we'll have time to discuss this in the WoT Profile meeting today, so I'll leave a couple of follow-up comments below.

> * the Think Link approach is specifically for the use case of one small device hosting a small number of _static_ Things, and for when you want to support self-discovery.  If you want to support a dynamic set of Things, then you really need an API to manage the set, notifications of updates, etc; that's the use case for the TDD service.   Please don't use the Thing Link approach for dynamic resources, that's not its use case (I think I said that somewhere in the text, also...).

OK, that makes sense.

> * I think it would be fine in Profiles to disallow directories in some cases.  However, the architecture means that the Thing can decide whether it wants to host a TDD service or not.  It's not forced to; if it is, please point out specifically what we would have to change in the spec.

The issue is not about what Things can do, but what Consumers must support. The point of a profile is to guarantee out-of-the-box interoperability between conformant Things and Consumers. This means that any feature which is optional for Things must necessarily be mandatory for Consumers. In other words, if Things can choose to host a Thing Directory then all Consumers must support Thing Directories. We could work around the problem by monkey-patching the WoT Discovery specification in the WoT Profile specification, defining that support for the well-known URI discovery mechanism is mandatory but excluding the Directory Description part. But it would be much cleaner if we could just reference introduction & exploration mechanisms directly.

Extending the well-known URI mechanism to support two different types of ThingLinks would mean that if the well-known URI mechanism was mandatory for the Core Profile, those features would have to be mandatory for _all_ conformant Consumers too, which further adds to the problem.

> Which gets to the last point, mixing in affordances into the Thing Link.

What we were suggesting was not adding affordances into a Thing Link, but using a Thing Description instead.

I'm not saying that you shouldn't extend the well-known URI mechanism*, only that doing so makes it even less feasible for it to be part of the Core Profile.

---

*I'm still not completely convinced by the necessity of ThingLinks at all, but that's a separate matter.

farshidtz (commented):
The PR is a bit drifting apart from the use case, which was about the limitation that well-known URLs don't allow setting base baths, making it impossible to expose multiple TDs from a single domain. See https://github.com/w3c/wot-discovery/issues/249

I now also think there is indeed a bit of overlap with the directory, in particular for the "external" relation. If the TD is external, it can probably self-describe itself using a separate well-known address. Or do we have another use case for this? 
If one or more TDs are related to each other, then another link relation can be chosen to define the nature of that relation.

Back to the original well-known issue and away from my initial standpoint: maybe it is better to allow returning a single TD or an array of TDs from well-known and leave Thing Link alone?

mmccool:
See comments on #208 for our resolution to this.  The conclusion was that rather than adding another mechanism for "simple" directories, we should make directories themselves easier to implement and use in simple cases.  So we are going to clean up the spec to make only "listing" mandatory (and registration optional).  Also, Issue #272 has been added to address the fact that we need to better describe what is expected of a *consumer* supporting discovery (e.g. they should be able to deal with at least the "listing" affordance of a directory).

mmccool:
We decided against the "just have .well-known return an array of TDs" as is breaks the invariant we decided upon: Introduction mechanisms always return (one or more) URLs that points at a (single) TD.  
Note that Intros can return multiple URLs (DID and DNS-SD being example) but .well-known can't, which is the root of the issue here.

mmccool:
Some of the comments above relate to improvements in the figure to make it clearer.  I'll see if I can address these as part of the resolution of Issue #272

Issue #238, opened 2021-11-22
Make XPath optional/informative


mmccool:
As mentioned in https://github.com/w3c/wot-discovery/issues/231, it was suggested that we remove XPath.  Let's discuss this and come to a consensus in this issue.

Some points so far:
1. Spec says XPath 3.1 which has a nicer syntax for arrays, etc. in JSON
2. There are a limited number of implementations of XPath 3.1 at present; note that LinkSmart supports only XPath 3.0 (go library; there is also a Java library (saxon), though).  UNIBO uses Node.js, no XPath library is available.
3. If two different TDDs use the same library to implement XPath queries, can we consider these two different implementations? - confirm with PLH
4. The above means that probably XPath queries will be "at risk" since we only have one implementation right now, and two are needed even for optional features.

mmccool:
My opinion is that we should keep it.  Arguments:
1. Optional, so does not constrain implementations
2. Functionally equivalent to JSONPath, provides simple query mechanism (not as pretty a syntax as JSONPath)
3. Testing issue can be resolved by having one more implementation based on Java/saxon for instance.

mmccool:
Discussion:
1. @k-toumura has experience using both XPath and SPARQL.  See https://github.com/w3c/wot-discovery/issues/231#issuecomment-946267769.  During the Nov 22 mtg, we discussed whether dropping XPath (and requiring use of SPARQL instead) would be sufficient for the use cases tested.  It seems so.
2. We do have two SPARQL implementations, so it is not at risk.
3. Likely that JSONPath will become standard and adopted by TDD implementations in practice; we just have to wait for Discovery 2.0 to make it official
4. Fact that SPARQL is relatively complex is perhaps not a big deal if it is hidden by query front ends, e.g. Node-RED discovery plugins.  A "basic" keyword clearly could be built on top of.
5. We *could* provide a bunch of examples of how to express semantics of equivalent JSONPath/XPath queries in SPARQL - to reduce the learning curve for developers.

mmccool:
Consensus:
1. Remove XPath (or make it informative, like JSONPath?) 
2. Focus on SPARQL; it will be the only normative query mechanism.
3. Provide a number of examples of how to map common JSONPath (and XPath?) queries into SPARQL.   Need to come up with some specific use cases, including at least the following:
     - filter TDs to those that only use HTTP
     - search for keywords in descriptions and titles
     - search for semantic types of things (e.g. look for "echonetAirConditioner")
[...]

PR #273, opened 2022-02-12
Things API with optional CRUD

### LINKED ISSUE #208: Read-Only Directories

### LINKED ISSUE #256: Small fixes on Creation
The first bullet point at 6.2.2.1.1 Creation says:

- A TD MUST be submitted to the directory using ...

I think it should be changed to

- A TD which is identified with an id attribute MUST be submitted to the directory using ...

I think it is a small incoherence since a TD does not necessarily have an id and this gives me the idea that a TD can lack an id field and be submitted to `/td/id` where the `id` inside the TD document is generated by the TDD.

Additionally, the text (also the TM of the TDD) does not say that the id inside the TD and the `id` in the `/td/id` must be the same. This should be made more obvious in my opinion. Also, what should the error be in this case? The HTTP semantics would say 400 but what the 400 means in this operation only includes ` Invalid serialization or TD.` which does not include mismatch of ids.

farshidtz:
This is WIP...

- renamed from Registration API to Things API to match the `/things` endpoint and justify optional "registration" interfaces; closes #208 
- renamed Notification API to Events API to match `/events` endpoint
- we already had Search API name for interfaces under the`/search` endpoint
- set listing to required in TD; related to #86 
- clarify creation assertion to explicitly mention where `id` comes from; closes #256

To Do:
- [ ] Convert text to assertions


directory.tm.json
```diff
@@ -9,9 +9,12 @@
     ],
     "title": "Thing Description Directory (TDD) Thing Model",
     "version": {
-        "model": "1.0.0-alpha"
+        "model": "1.0.0-beta"
     },
     "base": "{{DIRECTORY_BASE_URL}}",
+    "tm:required": [
+        "#/properties/things"
+    ],
```
farshidtz, line 17: The `things` property has [two forms](https://github.com/w3c/wot-discovery/blob/51c051368fc0015d3173164b4ae9ed0a0475a169/directory.tm.json#L55-L86). One is for array payload which is mandatory, another is for the collection payload which is optional. To define this required form, the TM should support setting of a form to required using a nested `required: true` field. It currently doesn't.

I am setting the whole property to required, assuming that it means at least one form should offer this interaction. I can also set to `#/properties/things/forms/0` but that means we are forcing the order of items in the forms array.

This was discussed in https://github.com/w3c/wot-discovery/pull/253#issuecomment-1004594898

farshidtz:
@RoboPhred, This PR closes your issue. I could not add you as the reviewer, but consider doing so.

index.html
```diff
@@ -876,41 +876,95 @@ <h3>Directory Service API</h3>
...
+                            order, or rejects the request and does not return the list.
+                            The server will return the list with the default order
+                            if and only if custom sorting is not requested.
+                        </li>
+                        <li>
+                            If the missing feature is the whole functionality at an API endpoint,
+                            the server may respond with 404 (Not Found).
+                            Alternatively, a server may respond with 501 (Not Implemented)
+                            to signify that it has placeholders in place for the given
+                            endpoint and would provide this functionality in future.
```
farshidtz, line 901: I wasn't sure if this should be an assertion. It won't help the clients anyway because it is optional for servers. Turning it into an assertion also means that we need to find two implementations who don't implement the feature and use those status codes!

directory.tm.json
relu91: TBH, I think that for now having only `things` property required is fine. I know that it is clearly depicting the whole picture but we have normative statements that cover that part. I would wait for the TD spec to catch up and deal with this kind of issue (I would say that the issue here is much more complex than just requiring that a derived TD must have a specific form).

index.html
relu91: I am not completely convinced about returning `501` either. Programmatically speaking having `404` or `501` result in the same output -> I cannot use this endpoint. About user, perspective knowing that the feature Might be implemented in the future feels like a temporary workaround. I would prefer chooseing just one of the two error codes.

relu91 (approved):

index.html
farshidtz: Removed 501 and converted 404 to an assertion, but recommended instead of optional.

index.html
```diff
@@ -2167,19 +2200,8 @@ <h2>Incremental Transfer</h2>
...
-            To do: add a
-            <a href="https://w3c.github.io/wot-thing-description/#thing-model">Thing Model</a>
-            (`directory.tm.json`)
-            that can derive to
-            <a href="https://github.com/w3c/wot-discovery/blob/main/implementations/example-directory-oauth2.td.json">this TD</a>
-            as an example.
-            The <a>TM</a> should describe all interaction affordances and indicate which ones
-            are required (i.e. mandatory by this spec). Moreover, it should use placeholders
-            for deployment-specific attributes.
-        </p>
```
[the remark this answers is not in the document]
farshidtz: Actually it isn't. We should still decide to define informative affordances: https://github.com/w3c/wot-discovery/issues/86#issuecomment-992591898

Added back the note with updates.

directory.tm.json
mmccool: OK for now with this resolution, but I'm wondering now if we should not have two different interactions here.  The expectation in TDs are that different forms are somehow equivalent, not for different purposes.  The scripting API, for example, also assumes that any form in an interaction will do the same thing as any other.

index.html
mmccool: OK.  I agree with only having one error code.  I am personally unsure why we don't make this mandatory but having it as a recommended is ok (probably best to not force an implementation to do anything at all if there is an error).

mmccool: The "may" should probably be MAY though and marked as an assertion with a span and id, however.  Are all errors written like this?  Do we have a general statement/assertion somewhere about errors that makes individual assertions unnecessary?

index.html
```diff
@@ -876,41 +876,96 @@ <h3>Directory Service API</h3>
                     incremental transfer of the payloads.
                 </p>
 
+                <p>
+                    The directory APIs include mandatory, recommended, and optional
```
mmccool, line 880: How do we relate these to RFC2119 assertions?  How do we capture them in the implementation report if they are not marked up in the text?

index.html
```diff
@@ -876,41 +876,96 @@ <h3>Directory Service API</h3>
...
+                    features. 
+                    Directories that don't offer recommended or optional features
+                    inform the clients about the absence of those features
+                    according to the following rules:
+                    <ul>
+                        <li>
+                            If the missing feature is to customize existing functionality of an
+                            API, the server will respond with 501 (Not Implemented)
```
mmccool, line 888: will -> MUST assertion?

index.html
```diff
@@ -876,41 +876,96 @@ <h3>Directory Service API</h3>
...
+                            HTTP status.
+                            For example, if a client requests a list of
+                            items with a particular sorting order, the server either
+                            respects that request and returns the list with the requested
```
mmccool, line 892: assertions?

index.html
```diff
@@ -876,41 +876,95 @@ <h3>Directory Service API</h3>
...
+                                    CRUDL (create, read, update, delete, and list) interfaces.</span>
+                            
+                                It is practical to implement only the interfaces to read and list
+                                if the directory serves a static collection of <a>TDs</a>.
+                                This is also useful for directories that intend to expose only
+                                retrieval operations over HTTP, and perform other interactions via non-HTTP protocols.
+
+                                <span class="rfc2119-assertion" id="tdd-reg-read-only-auth">
+                                    To expose read-only access, the directory MUST enforce access control
+                                    on create, update, and delete interfaces.</span>
```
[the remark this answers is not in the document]
mmccool: "other interactions via non-HTTP protocols" is not quite vague enough, and begs the question: "Why not describe it in the TD, then?".  I would make it even vaguer and say "other operations via out-of-band mechanisms."  This is meant to include means like internal (native) APIs (that may not even be network protocols).

index.html
```diff
@@ -974,11 +1041,13 @@ <h4>Creation</h4>
                             Error responses:
                             <ul>
                                 <li>
-                                    400 (Bad Request): Invalid serialization or TD.
+                                    400 (Bad Request): Bad user input such as
+                                    invalid serialization, invalid TD, `id` mismatch.
```
mmccool, line 1045: "TD, or "

index.html
```diff
@@ -1236,7 +1307,7 @@ <h4>Update</h4>
                         <h4>Deletion</h4>
                         <p>
                             <span class="rfc2119-assertion" id="tdd-reg-delete">
-                                A TD MUST be removed from the directory when an HTTP `DELETE` request is submitted
+                                A delete operation MUST be done using an HTTP `DELETE` request is submitted
```
mmccool, line 1310: strike "is submitted" or "MUST be done when an HTTP request is submitted"

index.html
```diff
@@ -2179,7 +2256,9 @@ <h1>Directory Service API Specification</h1>
             The <a>TM</a> should describe all interaction affordances and indicate which ones
             are required (i.e. mandatory by this spec). Moreover, it should use placeholders
             for deployment-specific attributes.
+            Informative affordances (JSONPath and XPath) should be defined as such.
```
mmccool, line 2259: JSONPath is definitely informative since there is no spec we can rely on.  XPath though can be normative, but optional (and SPARQL the same).

directory.tm.json
farshidtz: They are equivalent ways of getting the TDs. The only difference is payload and header formats.

index.html
```diff
@@ -2179,7 +2256,9 @@ <h1>Directory Service API Specification</h1>
...
         </p>
+
         The API of the directory is specified as a <a>Thing Model</a>.
         The Thing Model alone should not be considered as the reference
         to implement or interact with a directory.
```
mmccool, line 2264: A little unclear.  Should emphasize that all Thing Model constraints are normative, but there are *additional* constraints and assertions given in the body of the specification that also need to be followed by implementations.

mmccool (approved):
In the interest of time I think it is ok to go ahead and merge this, although there were a few places where I felt some improvements would be useful.  If there is time before the feature freeze an additional round of polishing may be a good idea.

index.html
farshidtz: @mmccool I think you are commenting on the outdated diff.

Regarding mandatory or recommended; as you said, forcing this means server have to implement it specifically that way. For example, a server may need to implement all those HTTP handlers and return 404.

index.html
farshidtz: There are appropriate assertions for those in the text.

index.html
farshidtz: Assertions are in the respective sections as explained below:
> The normative behavior is prescribed in the respective sections.

farshidtz: Assertions are in the respective sections as explained below:
> The normative behavior is prescribed in the respective sections.

index.html
farshidtz: This is just an example. Assertions are in the respective sections as explained below:
> The normative behavior is prescribed in the respective sections.

index.html
farshidtz: FIxed

index.html
farshidtz: Fixed

index.html
farshidtz: XPath is not normative. We had a consensus on this:
- https://github.com/w3c/wot-discovery/issues/238

index.html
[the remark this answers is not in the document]
farshidtz: Fixed

index.html
```diff
@@ -1546,7 +1579,7 @@ <h4>Management</h4>
                     </p>
                 </section>
```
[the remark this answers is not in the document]
mmccool: Have decided to remove.

directory.tm.json
mmccool: Discussion: Farshids thinks this is broken (maybe messed up as part of an earlier PR) and needs to be fixed.  McCool thinks at the very least that putting forms distinguished by different contentTypes in a single interaction is inconvenient for Consumers.  Farshid will create an issue and will investigate.

index.html
mmccool: ok, will let it go for now and review the record.

Issue #272, opened 2022-02-07
Consumer Process for Discovery


mmccool:
We need a clearer description of what the consumer has to do to implement discovery.  This should include:
- Managing and returning multiple introduction URIs
- Following multiple introduction URIs and merging multiple exploration results into a single *set* of TDs
- If Directories are encountered, optionally automatically following "listing" affordances and merging results... this may or may not be a good idea (e.g. the result list could be large, depending on the context), but the *option* to do it would be convenient in some contexts, e.g. Smart Home using a local hub for discovery
- NOT automatically following external links, e.g. TD Links support federated directories... which implies that if directories are discovered they may need to be included in the output from the discovery process, not elided
- Requiring that consumers that support discovery MUST support directories, at least to the point of using the (mandatory) "listing" affordance
- Consumers having the option to limit use of introduction mechanisms to those that limit scope to the LAN or to nearby devices (e.g. Bluetooth beacons, line of sight QR codes, etc).

Please comment below on the above or any other points that should be included.  The intention here is to capture the process as we intended, not to add significant NEW requirements or functionality.

mmccool:
I think the normative parts here are:
- Consumers MUST support multiple introduction results and multiple TD results
- Consumers MUST support at least "listing" affordance of any Directories TDs encountered
- Consumers MAY automatically expand "listing" affordances and add TDs to results
- Consumers MUST have a way to disable automatic expansion of discovery results if it is inappropriate
- Consumers MAY limit discovery to local context, e.g. to network segment or to proximate things, by limiting the introduction mechanisms used
- Consumers MAY use a limited subset of the introduction mechanisms

mmccool:
- This could all be considerably simplified if we dropped the automatic expansion of directory contents.  However, I think this was our original intention so that's what I'm trying to capture here, but making it optional so it does not become burdensome in cases where the directories have many TDs in them.
- Need to consider how to handle queries.  In the scripting API call, talked about making JSONPath (or something) a mandatory part of the discovery API, which would make it transparent whether it was supported remotely (if the directory supports it) or locally (if the directory does not and post-filtering is required).  Unfortunately JSONPath is not normative so we run into the same problem we had with making it normative on directories so for now I plan to not mention this or make it a requirement.  We should work on this in Scripting and wait for JSON Path to get finalized and deal with it in the next round...

mmccool:
[...]

Issue #271, opened 2022-02-03
Add a Directory to a Directory


benfrancis:
**User story:** As a user I want to add a directory of things to another directory so that I can manage multiple directories from a single larger directory.

**Use case:** A cloud service used for monitoring multiple buildings, each with their own Web of Things gateway. The user could add a whole building full of devices to the cloud service using the URL of its gateway's Directory Description.

Currently in order to achieve this each Thing Description from the gateway would have to be added to the cloud directory individually with a separate POST request.

Questions:
- Should the registration API allow the submission of a Directory Description to a Directory, in addition to individual Thing Descriptions?
- Should Directory Descriptions be stored directly in a directory such that a Consumer has to access the Things in the directory using its own `things/` endpoint, or should each Thing Description from the directory be added individually, perhaps with a link back to the original directory, e.g. with a "hosts" or "collection" link relation type?
- Related to the above, should directories be nestable (i.e. can you add a directory of directories to a directory?), or kept flat (Thing Descriptions from a Directory are stored individually, each with a link back to one or more other directories it may belong to)?

farshidtz:
This use case may be relevant: https://www.w3.org/TR/wot-usecases/#smart-campus

The registration API currently allows submission of any kind of TD, `discovery:ThingDirectory` is one of those types. Interested consumer would have to query /things endpoints recursively to retrieve nested TDs. I believe this is also what federated SPARQL queries would do:
> A WoT Thing Description Directory MAY implement federation in its SPARQL query API. If implemented, the SPARQL API MUST implement the SPARQL 1.1 Federated Query standard [[sparql11-overview](https://w3c.github.io/wot-discovery/#bib-sparql11-overview)].
https://w3c.github.io/wot-discovery/#search-semantic

You can design it either way. There are pros and cons in each. Here are my thoughts:

- scalability: TDs of devices from buildings in a districts or city will be hard to maintain in a single directory.
- access control: Nested directories can easily implement basic access control using network addresses. Flattened cloud directory would need an advanced mechanism to setup and enforce building-specific and global authorization. 
- ID uniqueness: Not a problem with random IDs, but hard to guarantee and scale in any other way.
- availability: Nested directories can still serve local consumer without internet access. Flattened cloud directories can be replicated and highly available.
- latency/networking costs: Cloud registration can be expensive on frequent writes. Recursive queries can be time consuming on frequent reads.
- maintainability: it is much easier to maintain a single cloud directory.

[...]

PR #274, opened 2022-02-13
Specify Consumer Discovery Process

mmccool:
Better describe and constrain what a *consumer* has to do to implement discovery.
Resolves Issue #272 

WIP: Do not merge until all checkboxes below are complete.

- [x] Update overview (Figure 1) to fix some typos (based on TD Link image, but reverting addition of TD Link path).  BTW, this is useful: https://github.com/jgraph/drawio-desktop
Also added notes to README explaining requirements for SVG files, and notes on how to export from drawio in particular (specifically: embed fonts).  This was because the SVG for the overview figure was broken (labels truncated), so I had to track down and figure out drawio, etc.
- [x] Add new section for Consumer Discovery Process
- [x] Add introductory text (copy/modify text from defunct PR #269 )
- [x] Add assertions (see discussion in Issue #272)
- [ ] Clean up intro text to align with proposed "simple directory" solution for multiple Things on one host
- [ ] Further update overview figure to indicate that expansion of contents of TD Directory is optional
- [x] <strike>Add</strike> Change Terminology definitions for <strike>D-Consumer and D-Client</strike> Discoverer

***
<a href="https://pr-preview.s3.amazonaws.com/mmccool/wot-discovery/pull/274.html" title="Last updated on Mar 7, 2022, 4:02 PM UTC (51b0070)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-discovery/274/f03fa88...mmccool:51b0070.html" title="Last updated on Mar 7, 2022, 4:02 PM UTC (51b0070)">Diff</a>


mmccool:
So, a couple of points to discuss:
- If Discovery is not mandatory for all Consumers, what entity do I make the assertions on?  I have defined a "D-Consumer" as a "Consumer that supports Discovery".  To discuss whether or not an assertion in Architecture saying "All Consumers MUST support WoT Discovery" to get rid of this.
- I have tried to keep the requirements for being a D-Consumer as low as possible.  In particular, I allow non-HTTP URLs, including "file", and specifically call out Consumers that support only a single Direct URL and only "file" to fetch that URL.  This still puts some light constraints on Consumers to be D-Consumers.
- Consumers can do what they want with Thing Links and Directories, including ignoring them.
- I describe the process of using a Think Link to refer to a Directory from other Directory (see #271).
- I make it mandatory to keep track of which Thing Links or Directories have been "expanded" but prohibit re-expansions (to avoid infinite loops when directories refer to each other).  To Do: what if I want to "refresh" the set of contents pulled from a directory?

mmccool:
We MAY want to be more particular about what it means for a Consumer to "support Discovery".  In particular I waffled a lot about whether requiring the ability to read a TD from an HTTP URL should be included, but it seemed overly exclusive.  CoAP consumers (without an HTTP client) may not be able to read from a Directory (which only has an HTTP API), but can do other things, such as use .well-known.  I have an ed note about that.

mmccool:
To be clear, Architecture only has the following assertion at this point, which makes Discovery optional:
* arch-discovery-should-use-standard: Whenever possible, the Discovery mechanisms defined in [[!WOT-DISCOVERY]] SHOULD be used to distribute TDs.

So Discovery is NOT mandatory for all Consumers.

mmccool:
@k-toumura makes a good point: the D-Consumer may not be a Consumer.  There are clients, like Toumura's Node-RED implementation, that may use Discovery but not itself be a Consumer.  So we need another term.  I want to avoid "Discoverer" (we have used it for something else) so maybe "Discovery Client" (D-Client for short).

mmccool:
@k-toumura I made some changes to the PR to address your point.  I add an additional definition, Discovering Client or D-Client, defined as "A client that supports Discovery".  I still have D-Consumers, but these are a subclass of D-Clients that can also interact with the affordances exposed by a TD etc.

There is some overlap: a D-Client *does* need to interpret certain parts of TDs, e.g. the @type (to figure out if it is a Directory or Thing Link) and also it needs to extract the URL from the form of the "listing" affordance in Directories.  It may also need to interpret some of the security information provided by Directory TDs.  But it's true it's not a "full" Consumer.

index.html
```diff
@@ -259,10 +259,198 @@ <h1>Architecture</h1>
...
+	In both cases, a Thing Link resolves a single URL into one or more additional URLs
+	which are fed back into the Introduction set, replacing the original URL pointing 
+	at the Thing Link.  A Thing Description Directory can also store Thing Links
+	which can redirect to other Thing Description Directories, allowing for a linked
+	directory structure.
+	</li>
+	</ul>
+    </section>
+    <section id="consumer-process" class="normative">
+        <h1>Discovering Clients and Consumers</h1>
```
farshidtz, line 308: I suggest keeping this under chapter 4. Architecture.

index.html
```diff
@@ -259,10 +259,198 @@ <h1>Architecture</h1>
...
+			DNS-SD, possibly in combination with mDNS. 
+			In this case the application 
+			code running in the D-Client
+			would not have to supply any additional information to invoke Discovery.
+		</li>
+		<li><span class="rfc2119-assertion" id="consumer-must-support-fetching">
+			A D-Client MUST support fetching a TD from least one URL provided as part of the Introduction process.</span>
+			In other words, not only must a D-Client accept URLs pointing at TDs, it must be
+			able to fetch that TD, for example by using a GET for an HTTP URL or by reading it from a file 
+			specified by way of a file URL.
```
farshidtz, line 376:
```suggestion
			specified by way of a file URI.
```
https://en.wikipedia.org/wiki/File_URI_scheme

index.html
```diff
@@ -259,10 +259,198 @@ <h1>Architecture</h1>
...
+    </section>
+    <section id="consumer-process" class="normative">
+        <h1>Discovering Clients and Consumers</h1>
+	<p>In this section we will describe the WoT Discovery process from the point of view of a client,
+	and what it means to say that a client supports WoT Discovery.
+	We will define a Discovering Client, D-Client for short, as an entity that supports the WoT Discovery
+	process.  Such a client may or may not be a Consumer.  A D-Client finds TDs, which are then passed
+	onto a Consumer for interpretation.  A D-Client itself may do some interpretation of TDs that it
+	fetches, however, in particular to examine the types of the TD it fetches to see if additional
+	fetches are required.  A Consumer may also be D-Client, in which case we will call
```
farshidtz, line 315: This sentence may need rewording.

index.html
```diff
@@ -259,10 +259,198 @@ <h1>Architecture</h1>
...
+	Every D-Consumer is also a D-Client.  The rest of this section refers only to D-Clients but all
+	assertions also apply to D-Consumers.
+	<p>
+	As described in section <a href="#architecture"></a>, WoT Discovery has two phases, Introduction
+	and Exploration. In the following sections we will give details on various Introduction and Exploration
+	mechanisms. 
+	However, <em>all</em> Introduction mechanisms result one or more URLs which can be used to fetch TDs by
+	a D-Client. 
+	Some of these TDs, however, may link to
+	TDs stored elsewhere (Thing Links) or describe web services that manage sets of TDs (Thing Description Directories, or TDDs).
```
farshidtz, line 353:
```suggestion
	TDs stored elsewhere (via Thing Links) or describe web services that manage sets of TDs (Thing Description Directories, or TDDs).
```

mmccool:
Things noted during Discovery mtg:
- SPARQL federation needs the URL of the SPARQL endpoint, not that of the TD for the TDD.  So the last sentence in the section needs revision
- "Discovering Clients and Consumers" is a confusing name, since "Discovering" can acts as either a verb or an adjective.  The latter is meant, but this is a relatively unusual English usage.  So I think changing this name to "Clients and Consumers" is best.

index.html
```diff
@@ -259,10 +259,198 @@ <h1>Architecture</h1>
...
+		</li>
+		<li><span class="rfc2119-assertion" id="consumer-must-support-fetching">
+			A D-Client MUST support fetching a TD from least one URL provided as part of the Introduction process.</span>
+			In other words, not only must a D-Client accept URLs pointing at TDs, it must be
+			able to fetch that TD, for example by using a GET for an HTTP URL or by reading it from a file 
+			specified by way of a file URL.
+			It is permissible to split the implementation of a D-Client into two parts, a "configurator" that
+			fetches/reads a TD and sets up a configuration during installation to be used by a 
+			run-time "executor" that cannot itself read TDs.
+			In this case, however, the "executor" is not by itself a D-Client.
```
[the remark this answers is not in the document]
mmccool: It is just an example, and these are names for these parts.  I could call them parts "A" and "B" instead.

index.html
```diff
@@ -259,10 +259,198 @@ <h1>Architecture</h1>
...
+	This is why in the following we allow any protocol to be used, even though
+	our Directory exploration mechanisms require HTTP.
+	</p>
+	<p class="ednote" title="D-Consumer">
+	Currently the WoT Architecture specification does not clearly say that Consumers MUST support Discovery.  
+	This is the reason for the introduction of the D-Client and D-Consumer terms.
+	</p>
+	</p>
+	We will use the term <a>D-Consumer</a> to refer to a WoT Consumer supporting WoT Discovery,
+	and D-Client as a more general term relating to any client supporting WoT Discovery.
```
farshidtz, line 343: The distinction between the two terms is not clear. I'm really not sure we need the two terms.

index.html
```diff
@@ -259,10 +259,198 @@ <h1>Architecture</h1>
...
+		<li><span class="rfc2119-assertion" id="consumer-fetch-links">A D-Client MAY fetch source TDs from the targets of the links in a Thing Link
+			described in its initial set of TDs and add them into the set of TD results.</span>
+			This only adds the results of fetching TDs from a Thing Link to the set of
+			results. These new results do not delete the original TD describing the Exploration mechanism. 
+		</li>
+		<li><span class="rfc2119-assertion" id="consumer-fetch-iteration">A D-Client MAY fetch additional TDs iteratively from any Thing Link or
+			Exploration mechanism described in its set of TDs and add them into the set of TD results.</span>
+			This only adds the results of fetching TDs from a Thing Link to the set of
+			results. These new results do not delete the original TD describing the Exploration mechanism. 
+		</li>
```
farshidtz, line 428: Is the second assertion referring to processing of nested links in TDs extracted from Thing Links? It is not very clear.

index.html
```diff
@@ -259,10 +259,198 @@ <h1>Architecture</h1>
...
+	using query filters when fetching TDs from a directory).
+	</li>
+	<li>
+	The second special case is a Thing Link.  This is also a Thing Description,
+	but rather than describing a Thing directly serves as a container for links to
+	other Thing Descriptions.  There are two categories of links supporting redirection
+	and sets.  Redirection allows for Thing Descriptions to be stored somewhere else
+	than at the exact URL provided by the Introduction.  Sets allows for single URLs to expand
+	into multiple Thing Descriptions without needing to use a Thing Description Directory.
+	Such sets are only appropriate for small static lists of Thing Descriptions.
```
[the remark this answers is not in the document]
mmccool: see updated version.  I believe I have resolved this now.

index.html
mmccool: I have not moved, but I have changed the name

index.html
mmccool: I think there are some cut-and-paste errors here.  Will try to address.

mmccool:
Fixed:
- Change name of section
- Clean up text in Architecture to remove mention of Thing Links for anything other than redirection
- Clean up text talking about SPARQL federation to make it clear the URL needed is not the one pointing at the TD itself, but at the SPARQL endpoint *given* in the TD.

index.html
```diff
@@ -259,10 +259,198 @@ <h1>Architecture</h1>
...
+			have already been fetched and avoid fetching duplicate results.</span>
+			This is to avoid infinite loops.  Note that Thing Description Directories can easily refer to each other,
+			for example, so fetching the TDs from a TDD, then fetching all those TDs will give another copy of the
+			original TDD's TD.  This second copy should NOT trigger further fetches.
+		</li>
+        </ul>
+	The above process supports a simple way to let Directories reference other Directories without duplicating their TDs: 
+	a Directory wanting to reference other Directories should include a Thing Link with a "describedby" relation to the
+	TD of the other Directory service.  Then the above process would expand the Thing Link to obtain the actual TD of the Directory, 
+	and then (optionally) use the appropriate Directory affordance to access the contents of the linked Directory.
```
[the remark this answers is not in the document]
mmccool: - I now have assertions saying external references do not *have* to be followed; indeed, dereferencing external links is a privacy risk
- More comments after I read the discussion in the above issue...

mmccool:
Introductions, Exploration, Directories etc. are mentioned in Architecture.  We could also put formal definitions into Terminology.  I definitely do need to talk about them in Clients.  Reversing the order of talking about Clients and Servers would resolve the issue too but would be less clear, I think (it would bury client behaviour down under a long laundry list of specific Introduction and Exploration mechanisms).

mmccool:
Some of the use cases for Thing Links (mentioned down much farther in the text) are bugging me.  I don't want to touch that part of the spec with this PR as it would cause conflicts, but at some point let's review those use cases.

farshidtz:
> I could also just introduce an explicit forward reference (e.g. "defined in Section xxx") for `ThingLink` and `ThingDirectory` values for `@type`.

That's also okay. But those types don't  currently have a section for themselves that are directly refencable.

index.html
```diff
@@ -259,10 +259,200 @@ <h1>Architecture</h1>
...
+		</li>
+		<li><span class="rfc2119-assertion" id="consumer-termination">A D-Client MAY terminate fetching additional TDs at any point or for
+			any reason.</span>
+	        </li>
+		<li><span class="rfc2119-assertion" id="consumer-any-order">A D-Client MAY fetch additional TDs by following links or fetching
+			additional TDs from Exploration mechanisms (e.g. TDDs) in any order.</span>
+			The D-Client can fetch additional TDs in any order from any Exploration mechanism and is not
+			required to use all of them.
+	        </li>
+		<li><span class="rfc2119-assertion" id="consumer-track">A D-Client MUST track (by id) which TDs describing links or Exploration mechanisms
```
relu91, line 432: Is `id` mandatory for ThingLinks and ThingDirectories? cause, if my mind is not failing me, it is not for generic TDs. For TMs loop cycle detection we are using the `href` of links as an ID.

relu91 (commented):
I tried to give my point of view about the current status which is already fine. Some additional points:
- a D-Client may fail to fetch some of the discovered URLs, should it abort the process or continue? is it the process still successful or not?
- I would also mention the fact that D-Client will probably need some "credential bootstrapping" to access TDs hosted at particular URLs. How should a D-Client behave if a fetch operation fails due to an `Unauthorized` response?

index.html
```diff
@@ -259,10 +259,200 @@ <h1>Architecture</h1>
                 alt="Discovery process overview" />
             <figcaption>Discovery process overview</figcaption>
         </figure>
-        <p class="ednote" title="Discovery Architecture Overview">
-        To do: an overview of the two-phase approach and its purpose, which is to support
-        controlled and authenticated access to metadata by authorized users only.
+	<p>
+	The WoT discovery process is designed using a two-phase approach.
+	In the first phase, one or more of a set of "Introduction" mechanisms may be used to 
+	generate a set of candidate URLs.  These URLs do not themselves contain metadata,
```
farshidtz, line 265: Is this the process of discovering a single Thing? Why more than one Intro mechanism is used? How do the clients decide between the candidate URLs?

index.html
```diff
@@ -259,10 +259,200 @@ <h1>Architecture</h1>
...
+	Introductions, however, do not include any security or privacy controls and
+	so should not provide metadata directly.  Instead, the provided URLs are followed
+	to "Exploration" mechanisms that will actually provide metadata but allow for
+	suitable security and privacy controls.
         </p>
+	<p>
+	Note that the discovery process can produce a <em>set</em> of URLs following 
+	its Introduction phase, even if only one Introduction mechanism is used,
+	and the final output after the Exploration phase
+	can be a <em>set</em> of Thing Descriptions.
```
farshidtz, line 278: Same as last comment. If the process is to discover one Thing, why is there a "set of URLs" after Intro, even if only one mechanism is used?

index.html
```diff
@@ -259,10 +259,200 @@ <h1>Architecture</h1>
...
+	All of the statements below about D-Clients also apply to D-Consumers.
+
+	These requirements are designed so that nearly any client that can fetch a single TD 
+	given a single URL over any protocol (including "file") can be said to support WoT Discovery. 
+	Of course, clients may support more powerful Discovery mechanisms, and some of these 
+	have their own requirements.  
+	For example, using WoT Thing Description Directories (TDDs) requires the use of HTTP, and so
+	are only available to clients that support this protocol.  However, use of HTTP (and Directories)
+	is not in fact a requirement of the WoT Discovery process: it is possible to discover multiple
+	TDs using just Introduction mechanisms.
```
farshidtz, line 321: Getting the TD is part of exploration. That is possible with or without a directory.

index.html
farshidtz: The components of a system can be named arbitrarily. I can argue that I need a Discovery Parser which parses the TD object incrementally. I can also argue that a TD is too large to pass around and should be transported and consumed incrementally (we have assertions for these).

index.html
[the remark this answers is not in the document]
mmccool: sure, making the names anonymous is reasonable

index.html
```diff
@@ -259,10 +259,200 @@ <h1>Architecture</h1>
...
+	Every D-Consumer is also a D-Client.  The rest of this section refers only to D-Clients but all
+	assertions also apply to D-Consumers.
+	<p>
+	As described in section <a href="#architecture"></a>, WoT Discovery has two phases, Introduction
+	and Exploration. In the following sections we will give details on various Introduction and Exploration
+	mechanisms. 
+	However, <em>all</em> Introduction mechanisms result one or more URLs which can be used to fetch TDs by
+	a D-Client. 
+	Some of these TDs, however, may link to
+	TDs stored elsewhere (Thing Links) or describe web services that manage sets of TDs (Thing Description Directories, or TDDs).
```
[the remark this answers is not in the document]
mmccool: Point noted, and will try to incorporate your changes, but I already merged another change from Farshid already here so can't apply your patch without some edits.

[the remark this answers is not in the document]
mmccool: I will have to look at the order of the text. Generally I want to expand acronyms on first use, but I'm not sure if this is the first use or not.

index.html
```diff
@@ -259,10 +259,200 @@ <h1>Architecture</h1>
...
+	The following assertions describe the responsibilities of a D-Client:
+	</p>
+	<ul>
+		<li><span class="rfc2119-assertion" id="consumer-must-support-intros">
+			A D-Client MUST support at least one Introduction mechanism.</span>
+			The simplest Introduction mechanism, Direct, simply provides a single URL of a target TD.  
+			This assertion results in different minimal requirements depending on
+			which Introduction mechanism is selected out of the several available.
+			For example, when Direct is used as the sole Introduction mechanism,
+			at a minimum a D-Client must be able to accept a single URL pointing at a TD.
```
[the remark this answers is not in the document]
mmccool: The Discovery spec should specify the minimum requirements for D-Clients.  It's already pretty loose ("any one Introduction") but we had a similar discussion with the Profile group about forcing clients to support HTTP.  A more realistic example than ftp is a Client that only supports CoAP.  Can it follow CoAP links to get TDs?  Yes, it won't work for Directories, but what about direct peer-to-peer access?

Let's discuss today.  Directories though need to be specifically for HTTP, so any use cases that need to use Directories won't work for other protocols.

index.html
```diff
@@ -259,10 +259,200 @@ <h1>Architecture</h1>
...
+			D-Client that only supports a single invocation of a Direct Introduction, 
+			directly specifying a single TD via a URL.
+		<li><span class="rfc2119-assertion" id="consumer-merge-intros">
+			A D-Client MUST be able to merge URLs resulting from 
+			multiple Introduction mechanisms, multiple results from a single Introduction mechanism,
+			and multiple Introduction invocations into a single set.</span>
+			Note that even if only one Introduction mechanism is supported multiple results might be 
+			produced.  This assertion states that the overall output of the Introduction phase is
+			a single set of URLs.  
+			The word "set" is used here in the mathematical sense: the results are unordered.
```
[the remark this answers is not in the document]
mmccool: Hmmm... true, duplicate removal is also implied by "set".  I think that is ok and what we want, but maybe it could be emphasized here.  I still think I want to explicitly state duplicate removal in an assertion and not depend on the mathematical definition of "set".

index.html
mmccool: Ack, good point.  A Thing Link or a Thing Directory without an id would be a pain.   I guess we could add "or href, if no id is available".

index.html
mmccool: Some Intro mechanisms can return multiple results.  Will discuss today.

index.html
mmccool: *Thought* I fixed this, let me checked (won't merge suggesting in case it screws up my other commits... but will take into account in my edits)

index.html
mmccool: Right, the problem is that "Discoverer" sounds like "an entity capable of doing Discovery" which is not, however, its definition.  We could
1. Use "Discoverer" rather than "D-Consumer" and "D-Client" and use another term for the current meaning (e.g. "Registerer")
2. Leave it as is.
3. Get rid of D-Client (since formally, since D-Clients need to process TDs, they are in fact D-Consumers).  However, definition of Consumer implies interactions with Things, including endpoint Things (not just directories, so @k-toumura would prefer different terms).

Consensus: do 1, and also use this to resolve 3 (say a Discoverer MAY be a Consumer, but not necessarily).

mmccool:
> I added several inline comments.
> 
> I also suggest moving the Thing Link and Thing Directory type definitions along with the ontology from [6. Exploration Mechanisms](https://w3c.github.io/wot-discovery/#exploration-mech) to somewhere under chapter 4. Architecture, because there are now forward referenced.

To do

mmccool:
I think the last use case is not really "describedby" but "extends" or similar.   I propose we  remove it:

A device intends to publish an entire [TD](https://w3c.github.io/wot-discovery/#dfn-td) which contains private and public parts: publish one [TD](https://w3c.github.io/wot-discovery/#dfn-td) (Thing Link) with only the public information referencing another [TD](https://w3c.github.io/wot-discovery/#dfn-td) which contains the full description.

index.html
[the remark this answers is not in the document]
mmccool: Pros of requiring HTTP: Discovery is more interoperable.
Cons: fragmentation into different verticals supporting different protocols.

So a Consumer only support CoAP would not be a Discoverer according to this spec.  CoAP also tends to go along with CBOR but we don't formally define that for TDs...

mmccool:
Update: just as we were closing the meeting Farshid said he would be ok in adding an assertion that Thing Links using the "describedby" relation should not have affordances.  So I'll add an assertion of that nature (such an assertion would not forbid Thing Links with other relation types from having affordances).  I would personally, however, like to extend this to say "should not have affordances or other metadata".  This would cover things like locations (or other random metadata using an extension), version info, security definitions, etc.  My goal is to allow a Discoverer to discard a Thing Link (using a describedby relation, anyway) once it is resolved.

index.html
```diff
@@ -259,10 +259,200 @@ <h1>Architecture</h1>
...
+	</p>
+	<p class="ednote" title="Minimum URL Requirements">Requiring the ability to read a TD from a URL over
+	HTTP would require even clients/consumers that only want to interact over CoAP, for instance, 
+	to implement an HTTP client to be able to support Discovery.
+	Another common way to configure a client/consumer is simply to read a (single) TD from a file.
+	There are also other valid ways to do discovery over CoAP, for example using CoRE-RD to discover
+	multiple TD links.
+	This is why in the following we allow any protocol to be used, even though
+	our Directory exploration mechanisms require HTTP.
+	</p>
```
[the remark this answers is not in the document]
mmccool: Huh, I misread this at first as "disagree"; but you actually "agree" with "not limiting to HTTP".  But as discussed below, I think limiting Discovery to HTTP will make things more interoperable.  I can see both sides, but I think dealing with CoAP/CBOR etc. right now is not entirely feasible.

index.html
mmccool: Ah, true.  Wording needs some tweaking.  Should say "without using Directories".

mmccool:
Did the following:
- Removed the third use case for Thing Links as discussed.  However, as I was doing so I re-read the other two and realized they also implied "non-pure" Thing Links, i.e. a link combined with other metadata and affordances.  Since that was my problem with the third use case, removing *just* it does not solve the problem.  I commented it out for now but did not actually delete the content...
- Rewrote section on Discoverer process, dealing with several of the issues noted in the discussion above.

mmccool:
Haven't implemented https://github.com/w3c/wot-discovery/pull/274#issuecomment-1047035147 yet.  It is probably still a good idea, but maybe we should handle it in a separate issue/PR, as we'll probably have to do some significant rewrites in that section, including removing/changing ALL the example use cases, which all seem to depend on partial TDs.

index.html
mmccool: WoT Discovery gives back a set of TDs.  The application then needs to decide what to do with them (Consume them all, pick one and Consume it, etc).  We do not define what clients do with the results of Discovery, just the Discovery process itself.

index.html
mmccool: reworded in update

index.html
mmccool: I changed this to just say that loops should be detected, not how.  Could also use full-result hashes, but I think this is an implementation decision, not a requirement.

index.html
```diff
@@ -259,10 +259,200 @@ <h1>Architecture</h1>
...
+			This only adds the results of fetching TDs from a Thing Link to the set of
+			results. These new results do not delete the original TD describing the Exploration mechanism. 
+		</li>
+		<li><span class="rfc2119-assertion" id="consumer-fetch-iteration">A D-Client MAY fetch additional TDs iteratively from any Thing Link or
+			Exploration mechanism described in its set of TDs and add them into the set of TD results.</span>
+			This only adds the results of fetching TDs from a Thing Link to the set of
+			results. These new results do not delete the original TD describing the Exploration mechanism. 
+		</li>
+		<li><span class="rfc2119-assertion" id="consumer-termination">A D-Client MAY terminate fetching additional TDs at any point or for
+			any reason.</span>
```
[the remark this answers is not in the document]
mmccool: I actually have a specific assertion that lets a Discoverer stop for any reason (for instance, they could simply have hit a time limit or a memory limit).  This will make discovery a bit inconsistent between implementations, but it would be anyway given the other variations allowed (e.g. types of intros).

index.html
```diff
@@ -259,10 +259,200 @@ <h1>Architecture</h1>
...
+		<li><span class="rfc2119-assertion" id="consumer-fetch-tdd">A D-Client MAY fetch additional TDs from any Exploration mechanism
+			described in its initial set of TDs (including, in particular, Thing Description Directories) 
+			and add them into the set of TD results.</span>
+			This only adds the results of fetching TDs from a Exploration mechanism to the set of
+			results. These new results do not delete the original TD describing the Exploration mechanism. 
+		</li>
+		<li><span class="rfc2119-assertion" id="consumer-fetch-links">A D-Client MAY fetch source TDs from the targets of the links in a Thing Link
+			described in its initial set of TDs and add them into the set of TD results.</span>
+			This only adds the results of fetching TDs from a Thing Link to the set of
+			results. These new results do not delete the original TD describing the Exploration mechanism.
```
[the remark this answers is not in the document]
mmccool: copy-and-paste issue.  Cleaned up in revision.

index.html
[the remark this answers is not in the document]
mmccool: added "unordered and unique"

index.html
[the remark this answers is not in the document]
mmccool: generally reorganized this part to try and make things flow better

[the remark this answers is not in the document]
mmccool: Think I will leave this for now, then we need another pass to clean up terminology.  I also use "Directory" in a lot of places where I should probably be using TDD.

farshidtz (approved):
Thanks, I think it is much clearer now. I still have some concerns but they aren't in assertions and can be addresses in future.

index.html
```diff
@@ -259,10 +259,207 @@ <h1>Architecture</h1>
...
+			for example, so fetching the TDs from a TDD, then fetching all those TDs will give another copy of the
+			original TDD's TD.  This second copy should NOT trigger further fetches. 
+			Note that tracking cannot depend only on the value of "id" fields in TDs since this
+			field is optional. 
+		</li>
+        </ul>
+	<p>
+	The above process supports a way to let Directories reference other Directories without duplicating their TDs: 
+	a Directory wanting to reference other Directories should include a Thing Link with a "describedby" relation to the
+	TD of the other Directory service.  Then the above process would expand the Thing Link to obtain the actual TD of the Directory,
```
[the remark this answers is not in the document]
mmccool: can we turn the above into an issue and deal with it in another PR?

mmccool:
Suggest we merge but create issues for the remaining problems and try to resolve them with focused PRs.

Issue #285, opened 2022-03-10
Security Consideration for Amplification Attacks


mmccool:
a * n * m model.
mitigations: limit total amount of data returned by introductions to be less than 3x request, following QUIC recommendations (so a < 3).  Disable observe (so n = 1) and multicast (so m = 1).  If a * n * m > 3, then require authentication before even "introduction".  Relates mostly to introduction mechanisms that can return multiple URLs, e.g. DID resolution, CoRE-RD, etc.

Also, introduction services running on a LAN should not accept requests from external sources on the internet.  Instead a directory (with access controls) should be used.

mmccool:
Closed since we merged PR #286.

PR #286, opened 2022-03-11
Add Amplification DDOS Security Consideration and Mitigations

mmccool:
Resolves issue #285 
- describe DDOS attack using open introduction mechanisms
- provide a set of mitigations
- Put after existing DOS security consideration
- existing DOS attack is not a DDOS attack, so rename, change id

***
<a href="https://pr-preview.s3.amazonaws.com/mmccool/wot-discovery/pull/286.html" title="Last updated on Mar 23, 2022, 5:10 PM UTC (88f5ea1)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-discovery/286/c997740...mmccool:88f5ea1.html" title="Last updated on Mar 23, 2022, 5:10 PM UTC (88f5ea1)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -2163,14 +2163,17 @@ img.wot-diagram {
             privacy issues, see the <em>WoT Security and Privacy Guidelines</em>
             specification [[?WOT-SECURITY]].
         </p>
-        <section id="security-consideration-ddos">
+        <section id="security-consideration-dos">
             <h2>Denial of Service</h2>
             <p>
                 Certain functions of the directory service,
                 in particular search queries,
                 may require significant resources to execute and this fact
-                can be used to launch DDoS attacks against
-            WoT Thing Description Directory services.
+                can be used to launch denial of service attacks against
+                WoT Thing Description Directory services.
+		In such an attack a WoT Directory would be overloaded
+		by requests from the attacker and unable to service other
+		requests.
 	    </p>
                 <dl>
                     <dt>Mitigations:</dt><dd>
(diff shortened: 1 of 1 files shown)


Citrullin (approved):

mmccool:
Discussed in Security call: Since TD has upgraded security considerations etc. to actual assertions, suggest we do that here also.  We should merge these first to resolve any merge conflicts among the two PRs in flight, then I can do that in a future PR.  Should discuss first in the Discovery call.

index.html
```diff
@@ -2187,6 +2190,72 @@ <h2>Denial of Service</h2>
...
+            </p>
+            <dl>
+                <dt>Mitigations:</dt>
+                <dd>
+		To mitigate this problem, the following is recommended:
+		<ul>
+		<li>Disable support for observe or similar extended result subprotocols
+		on Introduction mechanisms.</li>
+		<li>Introduction mechanisms should not respond to multicast
+		requests unless this is absolutely required by the protocol.</li>
```
JKRhb, line 2231: In the case of CoAP, there is a new RFC ([RFC 9175](https://datatracker.ietf.org/doc/html/rfc9175)) that seems to mitigate the amplification potential of multicast requests and other use cases.

JKRhb: Using the new RFC, the bullet point might be adjusted like so: 

```suggestion
		<li>In general, introduction mechanisms should not respond to multicast
		requests unless this is absolutely required by the protocol.
		In the case of CoAP, the recommendations made in [[RFC9175]] should be
		applied in order to mitigate risks of amplification attacks.
		</li>
```

I need to read a little bit more thoroughly thorugh the RFC myself, though.

JKRhb: Hmm, nevermind, I just read through [draft-mattsson-t2trg-amplification-attacks](https://datatracker.ietf.org/doc/draft-mattsson-t2trg-amplification-attacks/) where the same authors state that

> [f]or multicast requests, anti-amplification limits and the Echo Option do not really work unless the number of servers sending responses is known.

Maybe RFC 9175 could be cited elsewhere, though?

mmccool: yeah, they mentioned that too, you generally don't know the number of servers.  So, discouraging multicast is probably best, but it does mean that doing a multicast as part of an introduction (which is actually useful...) is a problem, but in an OPEN network (as opposed to a closed, trusted network).  Anyway, let me think about it, should probably cite that RFC and mention the issue that we don't know the number of servers, generally.

mmccool: Can still SAY what you propose, but temper a bit to say "If multicast is necessary..."

mmccool: ok, updated, pls take a look

JKRhb: Thank you, @mmccool, looks good! :)

j1y3p4rk (approved):
I agree on the changes.

Issue #291, opened 2022-03-16
Needed new assertion for discovery @context


AndreaCimminoArriaga:
When a TD is registered into a directory, the Directory adds the registration information. In order for this piece of the TD to be compliant with the JSON-LD 1.1 the directory must always insert the @context URI of the directory (otherwise the registration information is unbounded to an ontology). Currently, this feature is not present in the document (nor in the testing).

farshidtz:
How about when key/values are normalized?

E.g. `https://www.w3.org/2021/wot/discovery#dateCreated` without the context is equivalent to `created` with it.

For `@type`s, we accept also type URIs:

> A TD which describes a Thing Description Directory instance MUST use type ThingDirectory from the discovery context or URI https://www.w3.org/2021/wot/discovery#ThingDirectory.

> A TD which describes a reference to another TD MUST use type ThingLink from the discovery context or URI https://www.w3.org/2021/wot/discovery#ThingLink.

AndreaCimminoArriaga:
The following options are correct:
``` json
{
    "@context": [
        "http://www.w3.org/ns/td",
        "https://w3c.github.io/wot-discovery/context/discovery-context.jsonld"
    ],
    "@type": "DirectoryDescription",
    "title": "Thing Description Directory (TDD)",
    "version": {
        "instance": "1.0.0-alpha"
    },
   "registration": {
        "created": "2021-01-19T17:02:00Z",
        "modified": "2021-01-19T17:02:00Z"
    }
}
```
Alternatively:
``` json
{
    "@context":  "http://www.w3.org/ns/td",
    "@type": "DirectoryDescription",
    "title": "Thing Description Directory (TDD)",
    "version": {
        "instance": "1.0.0-alpha"
    },
   "https://www.w3.org/2021/wot/discovery#hasRegistrationInformation": {
        "https://www.w3.org/2021/wot/discovery#created": "2021-01-19T17:02:00Z",
        "https://www.w3.org/2021/wot/discovery#modified": "2021-01-19T17:02:00Z"
    }
}
```
Or:
``` json
{
    "@context": [
        "http://www.w3.org/ns/td",
      {"tdd":"https://www.w3.org/2021/wot/discovery#"}
    ],
    "@type": "DirectoryDescription",
    "title": "Thing Description Directory (TDD)",
    "version": {
        "instance": "1.0.0-alpha"
    },
  "tdd:hasRegistrationInformation": {
[...]

PR #292, opened 2022-03-16
PR for adding assertion of discovery @context

AndreaCimminoArriaga:
Added assertion related to issue #291

### DIFF
--- a/index.html
+++ b/index.html
@@ -878,7 +878,10 @@ img.wot-diagram {
                     stored in a directory. 
                     The following table lists the registration information attributes for
                     use within <a>TDs</a> that embed or reference the Discovery context. 
-                    Note that only an <a>Enriched TD</a> embeds the registration information.
+                    Note that only an <a>Enriched TD</a> embeds the registration information. 
+		    <span class="rfc2119-assertion" id="tdd-context-injection">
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -878,7 +878,10 @@ <h4>Registration Information</h4>
                     stored in a directory. 
                     The following table lists the registration information attributes for
                     use within <a>TDs</a> that embed or reference the Discovery context. 
-                    Note that only an <a>Enriched TD</a> embeds the registration information.
+                    Note that only an <a>Enriched TD</a> embeds the registration information. 
+		    <span class="rfc2119-assertion" id="tdd-context-injection">
+			An <a>Enriched TD</a> MUST contain in its @context the URI <code>https://w3c.github.io/wot-discovery/context/discovery-context.jsonld</code>
```
mmccool, line 883: I think it's time we picked and allocated more reasonable final URLs for all of these.  Note that we should also have security assertions noting that these are cached and built into implementations, and should NOT be fetched dynamically...

mmccool (approved):
good thing to update these, but I think our context URLs in general need to be updated, and I'd like to clean up these temporary github ones.  So we can merge this to get the text, but I think we should decide on and put into the doc more "final" URLs.

Issue #293, opened 2022-03-28
Upgrade Security and Privacy Considerations to Normative Sections


mmccool:
- make these normative, consistent with TD spec (and, will also be proposed for Architecture...)
- mark up appropriate statements under mitigations as assertions
- these would be manually asserted generally

Discussed in Security TF call, consensus was that making these sections normative is appropriate.

mmccool:
Please leave a comment here if you have an opinion on this.  In the meantime I will prepare a PR.  And yes, making these sections normative would add to the testing burden.  But on the other hand... it's probably the right thing to do, to insist that implementations mitigate known risks.

mmccool:
PR #295 should have closed this, closing manually.

PR #295, opened 2022-03-28
Make Security and Privacy Considerations Normative

mmccool:
Resolves Issue #293 

Some rewording was required to turn various statements into assertions.

Some HTML cleanup in the Performance Considerations (but that section is still informative).

***
<a href="https://pr-preview.s3.amazonaws.com/mmccool/wot-discovery/pull/295.html" title="Last updated on Apr 11, 2022, 1:37 PM UTC (ab0f82b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-discovery/295/12e8d17...mmccool:ab0f82b.html" title="Last updated on Apr 11, 2022, 1:37 PM UTC (ab0f82b)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -2212,8 +2212,6 @@ img.wot-diagram {
 		    </span>
 		    </li> 
                     </ul>
-                    In cases where queries are refused or aborted 
-                    appropriate error responses should be returned.
                     </dd>
                 </dl>
 	</section>
--- a/index.html
+++ b/index.html
@@ -2190,13 +2190,27 @@ img.wot-diagram {
 	    </p>
                 <dl>
                     <dt>Mitigations:</dt><dd>
-                    A WoT Thing Description Directory implementation should 
                     <ul>
-                    <li>Limit the number of queries per unit time from the same requestor</li>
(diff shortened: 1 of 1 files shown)


mmccool:
Notes:
- Can remove last sentence under DOS (for security, may not WANT to return error information, since then you'r just playing into the hacker's hands by wasting resources responding)
- 3x is motivated by a similar assertion in QUIC/HTTP3.  There is not a direct dependence on QUIC, but perhaps cite?
- *Could* also add a MAY to the 3x that introductions MAY accept extra "padding" data to allow for longer responses?  Read QUIC spec first, see if that is also allowed
- LAN self-discovery: reverse proxy does not need to be in the cloud, could be in gateway, as long as it has a public URL and can get a cert
- Under location tracking, may also want to regen the IP address as well (ipv6 has a feature to support this; RFC for this?) @Citrullin 
- Under query tracking, should we mention OpenID?

mmccool:
To Do:
- Delete "In cases where queries are refused or aborted appropriate error responses should be returned." in 8.1

Issue #308, opened 2022-05-12
Discovery document feedback


danielpeintner:
I read through the document and would like to share some of my findings/comments.
Please take it with a grain of salt since I am not English native speaker 🙃.

## Typos

I created https://github.com/w3c/wot-discovery/pull/307 with some typo fixes.

## Wording

Section [5. Discoverer Process](https://w3c.github.io/wot-discovery/#discoverer-process) has a paragraph that talks about two components but somehow does not not *properly* mention  the 2nd part...

Maybe it can be rephrased?

It is permissible to split the implementation of a Discoverer into two components, one that fetches/reads a TD and sets up a configuration during installation to be used by a run-time component that cannot itself read TDs.
->
It is permissible to split the implementation of a Discoverer into two components, one that fetches/reads a TD and **??? TWO THAT ???** sets up a configuration during installation to be used by a run-time component that cannot itself read TDs.

## Singular vs. Plural

Section [5. Discoverer Process](https://w3c.github.io/wot-discovery/#discoverer-process) again.

If multiple invocations of a single Introduction mechanism is supported
If multiple invocations of a single Introduction mechanism **ARE** supported

## Duplicate error codes

In [7.2.2.1.2 Retrieval](https://w3c.github.io/wot-discovery/#exploration-directory-api-things-retrieval), [7.2.2.1.3 Update](https://w3c.github.io/wot-discovery/#exploration-directory-api-things-update) , and  [7.2.2.1.4 Deletion](https://w3c.github.io/wot-discovery/#exploration-directory-api-things-deletion) there are duplicate error codes.

I don't think this is desired. Moreover, most of them are not sorted like 404, 401, 403

![grafik](https://user-images.githubusercontent.com/11281461/168085157-aabe01a3-2800-43f6-95df-5da7bee0394f.png)

mmccool:
Thanks Daniel for the review!

I'm prepping another PR that I hope will clean up Section 5.  Ben's review (Issue #299) mentions some of these points, esp in section 5.  We recently merged a [PR into Architecture](https://github.com/w3c/wot-architecture/pull/746) that updated the definition of "Discoverer" so that part of the general terminology cleanup needed is done at least.

Some of your other issues, e.g. about duplicate error codes, we probably want to have separate PRs for.  I expect I will have a PR that addresses at least some of these points on Monday, at which point we might want to break this issue and Ben's up into a set of smaller unrelated issues that we can have separate PRs addressing.

[...]

Issue #299, opened 2022-04-22
Feedback on latest Editor's Draft


benfrancis:
As with the Thing Description specification in https://github.com/w3c/wot-thing-description/issues/1324 I've carried out a full review of the latest Editor's Draft of the WoT Discovery specification prior to normative feature freeze, with a view to implementation in WebThings.

I hope this feedback is helpful. Sorry to include all of this in one big issue but there are a lot of minor typo corrections and editorial changes which don't necessarily justify their own issues. I'm happy to file separate issues for the bigger points if needed.

Most of my feedback is just editorial suggestions to make the specification more readable. Apart from some niggles regarding HTTP headers, content types and encryption requirements, my one big concern is the lack of a defined authentication mechanism for HTTP-based exploration mechanisms. In my view, until issue #135 is resolved, the discovery process is largely useless for ad-hoc discovery because the two step discovery process assumes that a Consumer already has authentication credentials for a device before it discovers it. I think this needs resolving prior to publication. There's a proposed solution in #135 but I'm waiting on feedback.

Reminder: There are quite a few remaining Editor's Notes and Issues which need resolving before publication.

# 3. Terminology 

- [ ]  The [wot-architecture] reference needs updating to [wot-architecture11] where the new terms are defined.
- [ ] The definition of the term "Discoverer" in WoT Architecture 1.1 appears to be inconsistent with the way the term is used in WoT Discovery 1.1. See https://github.com/w3c/wot-architecture/issues/738
- [ ] I think the specification could be made clearer by consistently using the following terms in the prose of the text:
  * Thing
  * Directory (aka Thing Directory or Thing Description Directory)
  * Thing Description
  * Directory Description (aka Thing Directory Description)
(Otherwise the distinction between a Thing Description (TD), Thing Directory (TD), Thing Directory Description (TDD) and Thing Description Directory (TDD) gets very confusing.)
- [ ] Note that there is currently not a definition of "Directory Description", though the term is used in several places.

# 4. Architecture

- [ ] Figure 1 seems to indicate that a directory is the only exploration mechanism, it isn't completely clear that self-description can be an exploration mechanism too. Also the section numbers are out of date.

# 5. Discoverer Process

[...]

Issue #135, opened 2021-03-12
Security bootstrapping during discovery


benfrancis:
**Problem**

There are a couple of sentences in the WoT Discovery spec which talk about only allowing authenticated entities to access TDs and TDDs.

> WoT Discovery has to allow authenticated and authorized entities (and only those entities) to find WoT Thing Descriptions satisfying a set of criteria, such as being near a certain location, or having certain semantics, or containing certain interactions.

 > The server SHOULD serve the requests after performing necessary authentication and authorization.

Currently security metadata is provided inside a Thing Description (for both things and directories), which causes a paradox:
1.  A consumer needs to parse the security medatadata in the Thing Description in order to know how to authenticate with the producer
2. A consumer can not access the security metadata in the Thing Description unless it is authenticated.

This is a problem with Thing Descriptions in general which is highlighted particularly well by the Directory Service API.

**Solution**

I propose a two step process by which a consumer can authenticate with the producer of a Thing Description (for both things and directories):
1. The first time a client fetches a Thing Description from the server, the server responds with a partial TD which contains only security metadata (and possibly the thing ID)
2. The client then performs the authentication steps necessary for (one of) the security scheme(s) described in the partial TD, before re-fetching the TD URL. Providing the authenticated user is authorised to access the TD, the server responds with the full TD.

First fetch:

```
{
    "@context": "https://www.w3.org/2019/wot/td/v1",
    "id": "urn:dev:ops:32473-WoTLamp-1234",
    "securityDefinitions": {
        "basic_sc": {"scheme": "basic", "in":"header"}
    },
    "security": ["basic_sc"],
}
```

Second fetch, with basic authentication provided in header:

```
{
    "@context": "https://www.w3.org/2019/wot/td/v1",
    "id": "urn:dev:ops:32473-WoTLamp-1234",
    "title": "MyLampThing",
    "securityDefinitions": {
        "basic_sc": {"scheme": "basic", "in":"header"}
    },
    "security": ["basic_sc"],
    "properties": {
        "status" : {
            "type": "string",
            "forms": [{"href": "https://mylamp.example.com/status"}]
        }
    },
    "actions": {
        "toggle" : {
            "forms": [{"href": "https://mylamp.example.com/toggle"}]
        }
    },
    "events":{
        "overheating":{
            "data": {"type": "string"},
            "forms": [{
                "href": "https://mylamp.example.com/oh",
                "subprotocol": "longpoll"
            }]
        }
    }
}
```

[...]

PR #311, opened 2022-05-15
Move error codes to Directory API's intro

farshidtz:
Related issues:
- #308
- #135
- #299

### DIFF
--- a/index.html
+++ b/index.html
@@ -1061,6 +1061,30 @@ img.wot-diagram {
                     the Problem Details error responses may omit the `type` field which defaults
                     to "about:blank" and set `title` to the HTTP status text.
                 </p>
+
+                <p>
+                    The APIs set the HTTP status codes as defined in 
+                    <a href="https://datatracker.ietf.org/doc/html/rfc7231#section-6">Section 6</a> of [[RFC7231]].
+                    The list of used error codes include (but are not limited to) the following:
+                    <ul>
+                        <li>
+                            400 (Bad Request): Invalid client input in body, query, or headers.
+                            This is accompanied by an appropriate response message.
+                        </li>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -1061,6 +1061,30 @@ <h3>Directory Service API</h3>
...
+                    The list of used error codes include (but are not limited to) the following:
+                    <ul>
+                        <li>
+                            400 (Bad Request): Invalid client input in body, query, or headers.
+                            This is accompanied by an appropriate response message.
+                        </li>
+                        <li>
+                            401 (Unauthorized): The request lacks valid authentication credentials.
+                            This is often accompanied by a `WWW-Authenticate` header containing
+                            information on how the client can authenticate and repeat the request.
```
farshidtz, line 1077: I use "often accompanied by `WWW-Authenticate` header", because I'm not sure how many Restful APIs actually do that, how many non-browser clients support it, and if it is relevant for all security schemes.

Anyway, I think the details of this is beyond the scope of the current specification because we don't discuss how the security should be implemented. It doesn't make sense to go too deep in just one piece of the workflow.

danielpeintner (approved):
You moved *all* error codes to a single place. LGTM and removes the same boilerplate over and over again...

index.html
mmccool: True, but we HAVE to do this for security bootstrapping.  Consolidation is good, and we care about security bootstrapping for WoT consumers specifically.

mmccool (approved):

mmccool:
- We should revisit the change in description of the 404 error, e.g. expected response, assertions related to that
- 401 response related to security bootstrapping, will have to expand upon this in a future PR.

PR #313, opened 2022-05-16
Security Bootstrapping 

### LINKED ISSUE #135: Security bootstrapping during discovery

### LINKED ISSUE #319: Figure 4 does not include Directory in Self-describing Thing class
Figure 4 does not show a Directory as being a subclass of Self-describing Things, even though the text immediately afterwards says, and I quote: "A directory may contain TDs and at the same time provide a TD and act as a self-describing Thing."

Also, the whole discussion of security bootstrapping and its proposed resolution (PR #313) would be cleaner if the only security bootstrapping we have to do is for self-describing things (including directories).

Therefore I would like to propose modifying Figure 4 to add a subclass arrow from Directory to Self-Describing Thing.

mmccool:
Resolves #135. Resolves #319.

- Add Security Bootstrapping subsection to Exploration overview
- Requires 401 responses from Exploration services
- Requires use of WWW-Authentication header in 401 responses
- Limits allowed authentication mechanisms in bootstrapping to Basic, Bearer, Digest, and <strike>OAuth</strike> OAuth2
- References S&P considerations in Architecture to govern use of TLS and access controls
- Cross-references discussion of 401 in exploration methods back to the security bootstrapping section (and since WWW-Authentication is now required, drops "often" language.)
- Removes assertions about TLS and HTTPS being required.  For reasons previously discussed a blanket assertion like this won't work.  Instead the WoT Architecture 1.1 is cited---also generally updated references to WOT-ARCHITECTURE to wot-architecture11.

The following is technically unrelated to bootstrapping, but I also upgraded an assertion about a self-describing device performing necessary authentication and authorization before providing access.  It was a SHOULD before which doesn't make any sense in combination with "necessary" so I upgraded it to a MUST.

***
<a href="https://pr-preview.s3.amazonaws.com/mmccool/wot-discovery/pull/313.html" title="Last updated on May 30, 2022, 2:48 PM UTC (3a7e5fc)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-discovery/313/81946d5...mmccool:3a7e5fc.html" title="Last updated on May 30, 2022, 2:48 PM UTC (3a7e5fc)">Diff</a>


index.html
```diff
@@ -787,6 +787,48 @@ <h4>`ThingLink`</h4>
...
                 </section>
             </section>
+            <section id="exploration-secboot" class="normative">
+                <h3>Security Bootstrapping</h3>
+                <p>The purpose of an exploration service is to serve TDs, but only after suitable
+                authentication, and only to authorized parties.  However, since we don't yet have
+                a TD when accessing an exploration service, we can't depend on the security metadata
+                held in TDs to know what kind of authentication and authorization information is needed.
+                Instead, we have to depend on existing security negotation support in HTTP to
+                bootstrap access, at least to the TD itself.
```
farshidtz, line 797: If TDs adds support for associating security definitions with links, we can Thing Links to split private and public TD parts.

farshidtz (changes requested):

index.html
```diff
@@ -810,11 +852,8 @@ <h2>Self-description</h2>
                 <dt>HTTP</dt>
                 <dd>
                     <p>
-                        <span class="rfc2119-assertion" id="self-http-secure"> 
-                            The HTTP-based self-description SHOULD be over HTTPS (HTTP Over TLS).
-                        </span>
                         <span class="rfc2119-assertion" id="self-http-access-control">
-                            The server SHOULD serve the requests after performing necessary 
+                            The server MUST serve the requests only after performing necessary
```
relu91, line 856: I'm concerned about this side-effect. I'm reading this as if every web thing is not allowed to publicly expose its TD, is that intended? Which are the downsides of allowing the possibility of exposing a TD to the public but allowing authenticated access for all (or a subset) of its affordances?

benfrancis: I agree it should be possible to have publicly accessible TDs which require no authentication/authorisation. I think this assertion does allow that by using the word _necessary_, but perhaps the language could be made clearer.

Another problem with this assertion though is that technically serving a 401 Unauthorized response is serving a request before authentication/authorization has taken place. The assertions might make sense if split up by different scenarios, e.g.

- If not authenticated where authentication is required, MUST respond with 401
- If authenticated but not authorized, MUST respond with 403 (maybe 404 in some cases?)
- If authenticated and authorized (where required), MUST respond with 200
- If the TD is not found, MUST respond with 404

index.html
```diff
@@ -787,6 +787,48 @@ <h4>`ThingLink`</h4>
...
+                <ul>
+                <li><span class="rfc2119-assertion" id="exploration-secboot-401">Exploration services 
+                MUST reply with an HTTP 401 response code if appropriate
+                authentication information has not been provided.</span>
+                </li>
+                <li><span class="rfc2119-assertion" id="exploration-secboot-auth">A 401 HTTP response 
+                from an exploration service MUST include a <code>WWW-Authentication</code> header
+                describing the required authorizations using one of the following IANA-registered
+                HTTP Authentication Schemes: Basic, Bearer, Digest, or OAuth.
+                </li>
```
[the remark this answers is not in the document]
benfrancis: @farshidtz wrote:
> These kinds of blanket statements aren't appropriate because it may make implementations impossible to comply.

Can you provide an example?

> Especially not when we are so close to publication with little time for testing.

I agree that isn't ideal, but I did highlight this problem a year ago in #135 and I don't really see how the specification can be published without fixing it, because it prevents discovery from working in the common use case where TDs are not public. Publishing public metadata for all the connected devices in a smart home (or in fact any private building) seems like a bad idea...

> I think we are trying to solve a TD limitation by something that isn't really equivalent nor protocol agnostic.

This is a problem that can't really be fixed inside the TD itself, because the problem is with accessing the TD which is part of the discovery process. The 401 solution is certainly HTTP-specific, but for the two exploration mechanisms defined in WoT Discovery 1.0 which both use HTTP, that seems sufficient.

Note that you did previously say:

>> Recommending that for all HTTP interactions is a good idea.
>> This is actually the mandatory approach for OAuth2: https://datatracker.ietf.org/doc/html/rfc6750#section-3

> The data in this header isn't always sufficient for a client to perform authentication.

I highlighted that as an assumption in #135:

>> This also assumes that the WWW-Authenticate header alone provides enough information for a client to authenticate with a directory without the need for any additional out-of-band information (except that provided as part of the authentication process like a user filling out a login form). Hopefully that is the case, but I don't know enough about each of the authentication schemes to be sure. 

If a `WWW-Authentication` header does not provide enough information for a client to be able to authenticate, then we may need to find another solution. However, if that's the case then how does it work in web browsers?

> Other concerns:
> 1. Is the SPARQL engine able to respond 401 with the required `WWW-Authentication`?
> 2. Do SSE clients read or understand `WWW-Authentication` header?
> 3. If the answer to 1 and 2 is no, the assertions should not be for the whole exploration APIs.

The security bootstrapping is only needed for accessing the TD itself. All of these other resources can have security metadata describing them inside the TD.

> 4. What if a server wants to hide a protected endpoint? From [Mozilla's description of HTTP Authentication](https://developer.mozilla.org/en-US/docs/Web/HTTP/Authentication#the_general_http_authentication_framework): "In all cases, the server may prefer returning a 404 Not Found status code, to hide the existence of the page to a user without adequate privileges or not correctly authenticated."

It seems reasonable to allow a 404 as an alternative to a 403 response.

> Is this a state of practice approach for RESTful APIs? Note that the workflows using this header usually involve a user and user-agent (browser).

I don't know. But if an authentication scheme prompts a user for authentication then a user needs to be prompted for authentication, there's not really any way around that. For authentication schemes which require user intervention, my assumption is that the API will need to be consumed by a UI which facilitates that interaction.

FWIW we've just started working on a WoT Directory client in WebThings Cloud and WebThings App which will both consume multiple WebThings Gateways, so we're going to need to find a solution to this problem. I can't guarantee that will be testable in time for the publication timeline though, so it would be great if other implementations could tackle this too.

index.html
benfrancis: If I've understood you correctly then splitting public and private TD parts into separate Thing Link and Thing Description resources would require that every (authenticated) directory actually has to serve two separate resources for every device - a public Thing Link and a private Thing Description. It may also need to serve both a public Thing Link and a private Thing Description for itself. How would that all fit into the prescriptive URL structure of the Directory Service API?

My original suggestion in #135 was that if a client requests a Thing Description without being authenticated, the server could initially respond with a partial TD which only contains security metadata (that could be the body of a 401 response). It could then re-request the same URL once authenticated. However, I think you pointed out that the security metadata in the Thing Description describes its affordances, not the TD itself.

Does the security metadata in a Thing Link describe the security requirements of accessing its links? I seem to recall that links are excluded from the scope of the security member, but I can't find text in the Thing Description specification which says that.

In order to avoid the need for two resources for every device in a directory, could an un-authenticated request on the URL of a Thing Description respond with a 401 response and contain a Thing Link in its body, linking to itself?!

```json
{
    "@context": [
        "https://www.w3.org/2022/wot/td/v1.1",
    ],
    "@type": "ThingLink",
    "id": "https://example.webthings.io/things/mything1",
    "links": [{
        "rel": "self",
        "href": "https://example.webthings.io/things/mything1",
        "type": "application/td+json"
    }],
    "securityDefinitions": {
        "oauth2_sc": {
            "scheme": "oauth2",
            "flow": "code",
            "authorization": "https://example.webthings.io/oauth/authorize",
            "token": "https://example.webthings.io/oauth/token",
            "scopes": [
                "/things/mything1:readwrite",
                "/things/mything1",
                "/things:readwrite",
                "/things"
            ]
        }
    },
    "security": "oauth2_sc"
}
```

If a `WWW-Authentication` header turns out to not contain sufficient metadata, the only other alternative I can think of is to define new semantics in the Thing Description specification for describing the security properties of the Thing Description itself, and it's probably too late for that.

index.html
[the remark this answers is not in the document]
farshidtz: @benfrancis, I think we are agreeing with each other. But the added assersions are not for the TD alone.

> 
> Can you provide an example?

SPAQRL and SSE, as mentioned.

> Note that you did previously say:
> 
> > > Recommending that for all HTTP interactions is a good idea.
> > > This is actually the mandatory approach for OAuth2: https://datatracker.ietf.org/doc/html/rfc6750#section-3
> 

Yes, I said recommending, not mandating.
We can re-mandate it for OAuth2, which would be a redundant assertion.
 
> > Other concerns:
> > 
> > 1. Is the SPARQL engine able to respond 401 with the required `WWW-Authentication`?
> > 2. Do SSE clients read or understand `WWW-Authentication` header?
> > 3. If the answer to 1 and 2 is no, the assertions should not be for the whole exploration APIs.
> 
> The security bootstrapping is only needed for accessing the TD itself. All of these other resources can have security metadata describing them inside the TD.

Well, this PR is mandating it for all HTTP endpoints. To make this only for the bootstrapping of the TD, it has to be part of self-describing only. Note that a directory that exposes its own TD is a self-decribing thing:
> A directory may contain [TDs](https://w3c.github.io/wot-discovery/#dfn-td) and at the same time provide a [TD](https://w3c.github.io/wot-discovery/#dfn-td) and act as a self-describing Thing.
> https://w3c.github.io/wot-discovery/#exploration-overview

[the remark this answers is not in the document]
mmccool: My intention with these assertions was to ONLY require them for the initial fetch of a TD from an exploration endpoint, not for everything else.  In these cases exploration services just act like "regular" web services except I also limit servers to only use a small subset of well-known and well-supported authentication mechanisms.  This means you can do the whole security-bootstrapping using ordinary web server (and client) libraries.

Good suggestion about allowing a 404 response.  I will also think about the wording of the assertions to make sure they are limited in scope to just the cases where we don't have a TD, e.g. for bootstrapping.  Once you have a TD you can use that instead (although the HTTP transactions still need to be consistent with it).

index.html
mmccool: I think serving partial TDs is basically a new security mechanism to negotiate access.  I'd rather not do that, since there is a perfectly good vetted negotiation mechanism in HTTP, which already has all kinds of support in libraries, etc.  Since (for now anyway) we are limiting ourselves to HTTP, this PR suggest that we simply use that.

index.html
mmccool: My point here is that you if say "SHOULD" and "necessary" together it's an oxymoron.  It would allow people to do the necessary authentication and authorization checks, find they fail, and STILL serve the TD, which is obviously not what we want.  If authentication and authorization fail, the resource should not be served, full stop.

mmccool: I'll have to think about the different scenarios.  Basically if authentication/authorization fails we don't want to serve the *resources*, but might still *reply* (with an error).  Let me think if I can tweak the wording to cover that.

mmccool:
I'm thinking I should restructure this a bit.  I did a few local repairs but since security bootstrapping is actually only needed for self-description, and first contact with a directory uses self-description, what I should do is merge the security bootstrapping section into self-description.  Otherwise I'm bumping into some redundancy, and case in point is the "necessary authentication and authorization" assertion, and complications like the fact that if you don't need a&a you can skip the 401, which is hard to say clearly with the current structure.

mmccool:
So there's another use case that came up, implied by the following sentence at the start of self-description: "The hosted <a>TD</a> may also be registered inside a <a>Thing Description Directory</a> as prescribed in [[[#exploration-directory]]]."   I think this is assuming an Introduction can point directly at a TD hosted by a directory.  Unfortunately, this would require security bootstrapping even for non-self-describing endpoints.  I want to disallow this case and only allow introductions to points at directories or self-describing things, full stop.  Then, as discussed, we only have to deal with security bootstrapping for self-describing things, since directories are also in that category.

However, I don't know if disallowing this violates some other assumptions people have been making.

mmccool:
Argh, I just realized Figure 4 does not show a "Directory" as a self-describing Thing, but it is, as we have been discussing.

mmccool:
Decide to hold off on restructuring while seeing if we can get the current structure to work.  I'm going to have to leave this alone for a while as I have other things to do, but will look at it again tomorrow.  If people can make another pass at it to see if I've addressed the concerns raised that would be helpful.  If not I have time tomorrow to do another pass of updates.

PS: I updated the figure but it did not rendering in a consistent way with the old one, probably due to PUML updates.  We can rerender all the figures to make them consistent when we are done.  Also, sorry for mixing fixes to multiple issues in one PR, but these two are related.

index.html
relu91: > My point here is that you if say "SHOULD" and "necessary" together it's an oxymoron. It would allow people to do the necessary authentication and authorization checks, find they fail, and STILL serve the TD, which is obviously not what we want. If authentication and authorization fail, the resource should not be served, full stop. If you want to serve a TD publically, then you can use nosec... IF it is permitted (does not have PII). In that case the "necessary" checks will always return "true" and the MUST condition is satisfied. So the assertion still allows for "public" TD distribution using nosec, as long as it's not PII.

Ah, I see your point, but I still would ask you to improve the sentence. I asked around a bunch of colleagues and all of them got the same impression. MUST together with "necessary" implies that you are always going to check for authentication and authorization and "no-check" feels that is not an option. I would like to see your comment above as an explainer to that assertion, maybe pointing to the fact that in WoT no_sec is still an authentication/authorization policy. Or maybe change the sentence to: 
>  The server MUST respond with the requested TD only after performing authentication and authorization if necessary.

relu91: > Regarding the ordering issue brought up by Ben I'll have to think about the different scenarios. Basically if authentication/authorization fails we don't want to serve the resources, but might still reply (with an error). Let me think if I can tweak the wording to cover that.

I would also agree that this is an issue. Explaining the different scenarios like @benfrancis is suggesting might also clarify the assertion too.

farshidtz:
> Argh, I just realized Figure 4 does not show a "Directory" as a self-describing Thing, but it is, as we have been discussing. In fact the text right after the figure _says_ the directory can "provide a TD and act as a self-describing Thing" so I _think_ this can be resolved by updating the figure.
> 
> [Created Issue #319](https://github.com/w3c/wot-discovery/issues/319). Also just updated this PR to address that issue.

This is because directories didn't have to be self-describing. It was an optional feature and something that isn't always required. For example, in a federated environment you may have many directories simply adding their TDs in a central directory. Why should every single directory have a public-facing server to serve its TD?

farshidtz:
> Update: One case I think this does break is people giving out "Direct" URLs for TDs hosted in directories. In fact we do this all the time, and I have even done it myself! But technically... maybe we will also have to extend security bootstrapping to "direct references to TDs hosted in directories". Which at least would justify having security bootstrapping in its own section, since we would then have to refer to it in at least two other places. I would rather not have to extend it further, though.

The security definitions for getting Things out of a directory is defined in the TD of a directory. The HTTP security bootstrapping can work as an alternative if it is fully compatible. I think we have concluded that they aren't as added by this PR: "there is only partial overlap with the schemes supported by TDs"

mmccool:
We need to discuss this today and quickly come to a consensus.  I realize this PR as it stands has some issues:
1. Directories being self-describing.  I don't think there is any practical problem in requiring this.  Any objections to making it so?
2. Which endpoints should support security bootstrapping? Currently the PR (after responding to Farshid's comments) only requires the "self-description" endpoints to support bootstrapping (implies that we choose "yes" to 1).  However, this means that direct access to, say, a TD stored in a directory by giving "things/{id}" in a URL would NOT (normatively, anyway) support security bootstrapping.  Since in a lot of the examples I have built myself I have used this *a lot*, I think this is a problem.  But then if we normatively say that these also need to support security bootstrapping, why not general queries, etc?  So another option would be to just require security bootstrapping *everywhere*, which was how Farshid originally interpreted my PR.

mmccool:
Farshid's suggestion that other endpoints can support bootstrapping "if they use security that is compatible", i.e. password, bearer, direct, oauth, or nosec, is also reasonable.  We just have to think about if this should be a SHOULD or a MUST.

benfrancis:
@mmccool I hadn't yet seen your latest replies when I replied so I'll address your new points below.

> 1. Directories being self-describing.  I don't think there is any practical problem in requiring this.  Any objections to making it so?

Is that not a problem for directories of directories? i.e. If a Directory Description is stored in a Directory, then a client needs to authenticate with the parent Directory in order to access the Directory Description of the child directory. This is the same problem as authenticating access to a Thing Description in a Directory except that the Thing Description happens to describe a Directory.

> Which endpoints should support security bootstrapping?

Only Thing Description resources need to support security bootstrapping. For simplicity it may make sense to restrict the set of allowed security schemes for the whole Directory Service API, but only the `/things/{id}` endpoint actually needs to provide WWW-Authentication headers for bootstrapping.

> But then if we normatively say that these also need to support security bootstrapping, why not general queries, etc? So another option would be to just require security bootstrapping everywhere, which was how Farshid originally interpreted my PR.

@farshidtz pointed out that some endpoints (e.g. /events using SSE and /search using SPARQL) might not actually be able to respond with WWW-Authentication headers (or at least existing implementations may not understand them).

benfrancis:
BTW, it's [`WWW-Authenticate`](https://httpwg.org/specs/rfc7235.html#header.www-authenticate), not `WWW-Authentication`.

mmccool:
OK, here is what I am seeing as a consensus (but we will discuss more in the Discovery call in an hour, pls comment if you disagree):
1. Directories should be self-describing.
2. Any URL that can return a TD MUST support security bootstrapping.  This includes not only self-discovery endpoints but "things/{id}" endpoints.  This covers all the URLs that would potentially be accessed by an Introduction (result URLs from an Introduction must point at a TD, which is also why directories must be self-describing...)
3. We'd use the WWW-Authenticate approach rather than trying to define our own negotiation mechanism.  This also means limiting ourselves to HTTP (which leaves self-description by CoAP or MQTT out of the scope of the current spec; it doesn't mean you can't do it, it just means it will not be normatively defined).

Optionally (for discussion):
- We can limit the supported authentication schemes for directories to the overlap of the WWW-Authenticate mechanism and schemes supported by TDs, which means the list already mentioned: password, bearer, digest, oauth, and nosec.  That would allow directory TDs to be consistent with security bootstrapping.  Technically though we would only have to make this limitation for endpoints that can return a TD (or whatever subset we say needs to support security bootstrapping).

mmccool:
I have made a number of updates to address the points raised in our discussion and feel that this PR is again ready for review.  In summary I:
- Reverted the exploration-class figure so that Directories are not a subclass of Self-Describing Things
- Reworded the Security Bootstrapping section to
     - discuss the use case as primarily applying to cases with humans in the loop and mention that for automation it would be better to pre-establish the security scheme to be used for exploration services
     - apply to any endpoint serving a TD, but not others
     - mention that other headers may be included in a response as well, e.g. as needed for OAuth
     - explicitly call out OAuth2 and limit security schemes that may be used to ones supported by TDs (this eliminates older known-to-be-insecure flows and also OAuth1)

mmccool:
OK, so the device flow is like the client flow in that the discoverer is also supposed to go to the authentication server first, and it doesn't really fit what we want (which is unfortunate, because it is *intended* for onboarding devices, but mostly devices that can display a code, like a TV) so I think we have to limit the security bootstrapping to the code flow - which technically requires user input as part of the process.

mmccool:
Have finished the things I needed to look into (OAuth2 flows, 302 Redirect response codes) and have finished updating the PR accordingly.  Also should mention I added some text better positioning the use cases for security bootstrapping.

index.html
```diff
@@ -787,36 +787,154 @@ <h4>`ThingLink`</h4>
...
+                A 401 HTTP response from an exploration service 
+                at an API endpoint intended to serve a TD
+                MUST include a <code>WWW-Authenticate</code> header
+                and any other headers
+                describing the required authorizations using one of the following IANA-registered
+                HTTP Authentication Schemes: Basic, Bearer, or Digest.</span>
+                For details of the requirements, the IANA registry should be consulted for
+                each of the above authentication schemes.  
+                <span class="rfc2119-assertion" id="exploration-secboot-oauth2-flows">
+                If a "302 (Redirection)" response code is provided during security bootstrapping,
```
farshidtz, line 877:
```suggestion
                If a "302 (Found)" response code is provided during security bootstrapping, 
```

index.html
```diff
@@ -787,36 +787,154 @@ <h4>`ThingLink`</h4>
...
+                <p>
+                There are Security and Privacy Considerations in [[wot-architecture11]] 
+                and [[wot-thing-description11]] that are
+                relevant regarding when authentication is required for access
+                to TDs and the use of secure transport.  
+                In summary, secure transport (e.g. TLS) is required for public services and strongly recommended
+                even on private networks (even if there is no authentication requirement, to
+                protect the confidentiality of queries), and <code>nosec</code> should only be considered 
+                in limited circumstances when no Personally Identifiable Information is 
+                present or can be inferred.
```
farshidtz, line 910: Same comment as before. Is "nosec" a defined security mechanism outside of the TD? If yes, a reference would be nice. If not:
```suggestion
                protect the confidentiality of queries), and serving requests without authentication and authorization should only be considered 
                in limited circumstances when no Personally Identifiable Information is 
                present or can be inferred.
```

index.html
```diff
@@ -852,7 +970,18 @@ <h2>Self-description</h2>
...
+                                all cases other than OAuth2.
+                                Information on what credentials are required will be included 
+                                in a <code>WWW-Authenticate</code> header.
+                            </li>
+                            <li>
+                                302 (Redirection): Redirect to an authorization server.
+                                This should be used only in the case
+                                of initiating the OAuth2 <code>code</code> flow, 
+                                in particular for for security bootstrapping. 
                             </li>
```
farshidtz, line 985: This is the list of error responses.
```suggestion
```

index.html
```diff
@@ -1080,14 +1212,23 @@ <h3>Directory Service API</h3>
                     <a href="https://datatracker.ietf.org/doc/html/rfc7231#section-6">Section 6</a> of [[RFC7231]].
                     The list of used error codes include (but are not limited to) the following:
                     <ul>
+                        <li>
+                            302 (Redirection): Redirect to an authorization server.
+                            This should be used only in the case
+                            of initiating the OAuth2 <code>code</code> flow, 
+                            in particular for for security bootstrapping on endpoints
+                            that provide access to TDs.
+                        </li>
```
farshidtz, line 1221: Not an error code. Why can't 302 be used for purposes other than Oauth2?
```suggestion
```

index.html
```diff
@@ -787,36 +787,154 @@ <h4>`ThingLink`</h4>
...
+                that can be used for access.
+                To enable security bootstrapping on a wide variety of devices as well
+                as on browsers, the following constraints should be observed:
+                </p>
+                <ul>
+                <li><span class="rfc2119-assertion" id="exploration-secboot-401">
+                After initial contact with an exploration service using the URL provided
+                by an introduction mechanism, the exploration service
+                MUST reply with either an HTTP "401 (Unauthorized)" response code 
+                or and HTTP "302 (Redirection)" response code (OAuth2 only) if appropriate
```
farshidtz, line 853:
```suggestion
                or HTTP "302 (Found)" response code (OAuth2 only) if appropriate
```

index.html
```diff
@@ -787,36 +787,154 @@ <h4>`ThingLink`</h4>
...
+                authentication information has not been provided but access can be granted
+                when it is.</span>
```
farshidtz, line 855: Again, having a MUST here limits us to a few mechanism that we know of. 

1. The oauth2 spec allows other redirection mechanisms: https://datatracker.ietf.org/doc/html/rfc6749#section-1.7
2. I've mentioned mTLS before (where the client authenticates using a certificate in TLS negotiation) which is very secure, supported by browsers and HTTP clients, and very relevant for human to machine (e.g. with a smart card) and machine to machine (certificate issued by vendor and installed on IoT devices) interaction. I'm not sure that 401/302 covers it. 
https://www.cloudflare.com/learning/access-management/what-is-mutual-tls/
https://medium.com/@sevcsik/authentication-using-https-client-certificates-3c9d270e8326

farshidtz (changes requested):
Thanks @mmccool for the updates. I think the security bootstrapping is much better now, addressing a few common security mechanisms. I've added some comments.

index.html
```diff
@@ -787,36 +787,154 @@ <h4>`ThingLink`</h4>
...
+                <p>
+                In cases when exploration services are being used to
+                automate system management it would be best to pre-establish what credentials (and
+                authentication mechanisms) are needed to access the relevant exploration services
+                and security bootstrapping would not be required.  
+                </p>
+                <p>Security bootstrapping is also only (potentially) necessary on the <em>first</em> access to
+                a TD. Once a Discoverer has determined what credentials and authentication mechanism
+                are required to access a particular exploration service, they can retain them
+                and use them for future accesses.
```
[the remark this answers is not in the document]
mmccool: sure, that is a reasonable observation and fix

index.html
```diff
@@ -787,36 +787,154 @@ <h4>`ThingLink`</h4>
...
+                Instead, we have to depend on existing security negotation support in HTTP to
+                bootstrap access, at least to the TD itself.
+                </p>
+                <p>
+                Most of the HTTP negotiation processes assume there is a human user in the loop,
+                but this is also appropriate for WoT Discovery, since this problem will
+                typically occur when a user is trying to access a public WoT service or perform
+                integration of a new device. In this case the purpose of negotiation is to
+                inform the user of any steps they need to take to obtain credentials to access
+                the service.
```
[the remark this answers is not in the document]
mmccool: I guess I was thinking Oauth2 flows that in general do provide some guidance.  But yes, in general, a system could let you know it needs a password without telling you how to get one.  I will try to reword.

index.html
```diff
@@ -787,36 +787,154 @@ <h4>`ThingLink`</h4>
...
+                a TD. Once a Discoverer has determined what credentials and authentication mechanism
+                are required to access a particular exploration service, they can retain them
+                and use them for future accesses.
+                </p>
+                <p>
+                <span class="rfc2119-assertion" id="security-bootstrapping-endpoints">Security
+                bootstrapping MUST be provided on any HTTP endpoint that serves a TD.</span>
+                This applies not only to endpoints serving TDs from self-describing Things,
+                but also to endpoints such as <code>things/{id}</code> in <a>WoT Thing Description Directory</a>
+                exploration services.
```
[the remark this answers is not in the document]
mmccool: Yes, it was an annoying forward reference, your fix is good.

index.html
```diff
@@ -787,36 +787,154 @@ <h4>`ThingLink`</h4>
...
+                Note that if the exploration service does not want to provide access for
+                some reason, it can ignore the request or reply with another code such as 404 or 403.
+                Also, if no authentication is required, equivalent to the 
+                <code>nosec</code> scheme, then the 
+                system can reply immediately with the requested TD as if authentication information
+                were provided.  
+                Bypassing authentication however (that is, effectively using the <code>nosec</code>
+                security scheme to provide the TD) is only
+                appropriate if the TD served as a response does not contain and cannot be used to infer 
+                Personally Identifiable Information.
```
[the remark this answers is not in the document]
mmccool: I was using "nosec" as a shorthand for "no authentication" but it's not really appropriate, so I *should* try to avoid it.  Your suggestion is acceptable.

index.html
```diff
@@ -787,36 +787,154 @@ <h4>`ThingLink`</h4>
...
+                MUST include a <code>WWW-Authenticate</code> header
+                and any other headers
+                describing the required authorizations using one of the following IANA-registered
+                HTTP Authentication Schemes: Basic, Bearer, or Digest.</span>
+                For details of the requirements, the IANA registry should be consulted for
+                each of the above authentication schemes.  
+                <span class="rfc2119-assertion" id="exploration-secboot-oauth2-flows">
+                If a "302 (Redirection)" response code is provided during security bootstrapping, 
+                the OAuth2 <code>code</code> flow MUST be used,
+                with access credentials eventually being represented with bearer tokens.
```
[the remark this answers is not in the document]
mmccool: It's true that OAuth2 does not say that 302 is the only way to do this but I was hoping to limit the complexity of the Consumer by limiting the options; that was why I added this assertion.   Leaving it completely open (as the OAuth2 spec does) will lead to interoperability problems.  But I agree with the later changes to remove 302 from the error responses, since it is not really an error.  If there are other ways to specify redirections I'd rather just include them, e.g. " redirection can be accomplished with 302 OR X".

index.html
mmccool: I think the issue with mTLS is privacy, so it's not a silver bullet.  I'm not against including but would like to discuss it separately, as it may also impact the TD spec.

mmccool:
fixed all the "easy" fixes for which suggestions were supplied (all were merged, in summary) but have to think about the rest of them.  I'll try to reword the "negotiation" statement soon.  As for insistence on using 301 for OAuth2 redirection, I was trying to limit the number of options to simplify implementations.  Technically OAuth2 could allow other redirection mechanisms, true, but 302 seems to be what is done in practice.   If you are aware of other redirection schemes that are in common use I could add them but would like to have a finite list for implementability.

mmccool:
Reworded negotiation sentence to "In this case the purpose of negotiation is to provide guidance on what credentials are needed to access the system."

I feel this is a true statement; negotiation tells the user they need a password (for example) but not where to get one.  The user will have to figure that out on their own doing human-level things.

mmccool:
Let's leave the 302 issue to discuss in the meeting.  Options are:
1. Take it out.  This means "any form of redirection" as stated in the OAuth2 spec.  But I have concerns this leaves a hole for interoperability problems on lightweight consumers.
2. Leave it as is.  Then only 302 would be supported.
3. Add other mechanisms in common use (if there are any).    I plan to do a little research to see if there are any.

mmccool:
I looked into the redirection issue a bit and found that while most examples use 302 (Found) for redirection, the OAuth2 spec itself uses 303 (See Other) in its examples.   Code 302 is supposed to be for "temporary redirect" which makes sense.  Other 300 codes do not make sense; they are for permanent redirection, proxies, multiple responses, and the like.    The difference between 302 and 303 is that 303 requires the method to be changed to GET, in 302 it is optional.  There is also 307 which never changes the method, and is meant for dealing with pages that are temporarily unavailable (reloading gives you the same method).

303 does make make the most sense but 302 can also be used if the target responds the same way to all methods.  So I think what I should do here is change the text to "302 (Found) or 303 (See Other)" (at least). 

Here is a reference: https://developer.mozilla.org/en-US/docs/Web/HTTP/Redirections

farshidtz:
If the "code" flow meant to have a human in the loop, why we need interoperability at the API level? A redirection shall happen, one way or another, and the user (with the help of the browser) will be prompted to login. 

There are a few HTTP redirection codes, but 302 seem the most appropriate for Oauth2 code flow. Status 307 is only appropriate of the resource request is with GET request and the redirected request expected to have the same method.

Outside of a RESTful API, with JS as you mentioned, [HTML redirect](https://www.w3docs.com/snippets/html/how-to-redirect-a-web-page-in-html.html). It may be different in mobile apps.

index.html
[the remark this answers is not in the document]
farshidtz: I replied to your comments in the PR. Basically questioning why we need interoperability here.

Regardless, IMO, the assertion isn't really mandating what you intend but instead saying that a 302 means we need to do Oauth2 code flow, assuming 302 is exclusively used for Oauth2.

I still think the assertion isn't needed, but if you insist:
```suggestion
                Security bootstrapping with OAuth2 `code` flow MUST use the "302 (Found)"
                response to redirect the client to a URL supplied in the Location header.
```

benfrancis:
What I'm hoping to achieve here is out-of-the-box interoperability such that given the URL of a Thing Description (whether that Thing Description is hosted in a Directory or a by a self-describing Thing), any WoT Consumer from any vendor is able to prompt a user for credentials to access that Thing Description. If support for a standard security bootstrapping mechanism (perhaps more usefully described as "TD authentication") is made optional, then in the absence of a standardised mechanism for a Consumer to authenticate, Consumers will require custom code for the security schemes implemented by particular Directory implementations. In practice this is likely to mean that only Consumer and Directory implementations from the same vendor will work together, or that Consumers will require per-Directory customisation.

I think implementation experience may help improve our shared understanding of the feasibility of out-of-the-box interoperability here. My current impression is that if security bootstrapping is made optional, such interoperability won't be possible. That will drastically reduce the usefulness of WoT Discovery, since only Consumer & Directory implementations from the same vendor are likely to work together out of the box.

My hope had been that the WoT Discovery specification on its own would be prescriptive enough to enable out-of-the-box interoperability. If it isn't then we may need to add a Discovery section to the HTTP Profile which enforces additional constraints on directories and self-describing Things, but we're running out of time to define something that complex on top of a WoT Discovery specification which is not yet stable itself.

mmccool:
Ben's point is something I also worry about but I also don't want to weaken security generally.  I think we should merge this now, and think about other ways to address this.  I think making security bootstrapping mandatory in some profiles would be one way.  Another way would be to define an onboarding process and then allow for security bootstrapping to be disabled after onboarding (but I didn't want to shoehorn an onboarding spec into discovery).

So I think we should merge this so we have something for wide review, but then try to resolve the issue with some additional discussion.

farshidtz:
@benfrancis 
> If support for a standard security bootstrapping mechanism (perhaps more usefully described as "TD authentication") is made optional, then in the absence of a standardised mechanism for a Consumer to authenticate, Consumers will require custom code for the security schemes implemented by particular Directory implementations. In practice this is likely to mean that only Consumer and Directory implementations from the same vendor will work together, or that Consumers will require per-Directory customisation.

This is made optional but it is still normative and most parts are mandatory per HTTP and OAuth2 specs anyway. Many implementations will probably end up respecting it.

I suggested originally to make it a recommendation (SHOULD) but according to @mmccool some discussion in security TF led to making it optional.

PR #318, opened 2022-05-19
UTF-8 Constraints and Language Negotiation

### LINKED ISSUE #279: Validation section internationalization
In https://www.w3.org/TR/wot-discovery/#validation we define additional parameters for describing errors in TDs. Those parameters might contain natural language.

### LINKED ISSUE #281: Assert use of UTF-8
In order to support localization, API should return all text using UTF-8.
* May be implicit in encoding of JSON, to check.  (JSON says "should" and permits UTF-16, etc; we probably want to upgrade that a MUST).
* May be implicit in definition of Problem Details, to check.
* There IS an assertion that TDs must be encoded using UTF-8.   So registration, listing, etc. will not have to change the encoding of TDs.
* An assertion in Discovery means that other text, including queries, enriched text, errors, etc. also have to be encoded in UTF-8.

TO DO: we may need to add an assertion to mandate UTF-8 for all interactions and API data/responses.

### LINKED ISSUE #304: Missing feature to filter by language
See [TD Issue #1413](https://github.com/w3c/wot-thing-description/issues/1413)

### LINKED ISSUE #316: Specify Sorting Algorithm
While filling out the i18n questionnaire I ran into a question asking if the spec (or its implementation) sorts text. It does; we currently have some capabilities to sort by id but ALSO "other fields", which could be text in various languages. We probably should cite some recognized mechanism to sort UTF-8 strings. I would like to suggest using a pure lexicographic sort on the binary encoding of the UTF-8 string, since doing it in a language-specific way will be hugely complicated as each language will have their own rules, we will have to worry about how to compare strings from different languages, etc.
Since we need to do sorting on what could be very small devices I think using UTF-8 lexicographic sorting is the only feasible option.
 I also wonder how you would sort if the field you select happens to be something other than a string number or URL, i.e. a JSON object or array... disallowing that is probably a good idea.

mmccool:
Resolves #316.  Resolves #281.  Resolves #304.  Resolves #279.

- Requires UTF-8 encoding of Problem Details
    - In Problem Details in validation
    - In Problem Details for directory API
    - ... Problem Details not required for self-description... huh.
- Allows language negotiation using Accept-Language header for both TDs and errors
    - originally I was only planning this for errors, but it's inconsistent to not also honor it for TDs
    - you need the Accept-Language header on POST and GET requests that normally would register and fetch TDs, so if I do a GET on the "things" endpoint with Accept-Language set to some value, logically it would apply to the valid results, not just to errors...
    - There was an assertion for "Accept" but is for Content-Type, not language: "The server MAY provide alternative representations through server-driven content negotiation, that is by honouring the request's Accept header and responding with the supported TD serialization and equivalent Content-Type header."
    - I also noticed there was an assertion for this in self-description but not for directories, so while I was at it I added a similar assertion for directories
- Defines sorting order to be lexicographic UTF-8
    - Simple enough for small devices
    - Language-specific sorting rules for ALL languages just not feasible on smaller machines
- Disallows sorting on fields that are not basic types
    - Don't know if we need to define this, or if there is a JSON standard we can reference
    - But generally the spec allows any implementation to opt out of sorting on anything but ids anyway

***
<a href="https://pr-preview.s3.amazonaws.com/mmccool/wot-discovery/pull/318.html" title="Last updated on May 23, 2022, 2:20 PM UTC (752d730)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-discovery/318/3afc3df...mmccool:752d730.html" title="Last updated on May 23, 2022, 2:20 PM UTC (752d730)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -827,18 +827,16 @@ img.wot-diagram {
                             Content-Type header, and the <a>TD</a> in body.
                         </span>
                         <span class="rfc2119-assertion" id="self-http-alternate-content">
-                            The server MAY provide alternative representations through
+                            The self-description server MAY provide alternative representations through
                             server-driven content negotiation, that is by honouring the 
                             request's Accept header and responding with the supported
                             TD serialization and equivalent Content-Type header.
                         </span>
                         <span class="rfc2119-assertion" id="self-http-alternate-language">
-                            The server MAY provide modified TDs using a different
-                            default language after server-driven content negotiation, 
+                            The self-description server MAY provide modified TDs or error responses 
+                            using a different default language after server-driven content negotiation, 
                             that is by honouring the request's Accept-Language header.
                         </span>
-                        As will be discussed later, the Accept-Language header may also influence
-                        the language used for error responses.
                     </p>
                     <p>
                         <span class="rfc2119-assertion" id="self-http-head">
@@ -1151,6 +1149,19 @@ img.wot-diagram {
                         </li>
                     </ul>
                 </p>
+
+                <p>
+                  <span class="rfc2119-assertion" id="tdd-http-alternate-content">
+                     A Directory server MAY provide alternative representations through
+                     server-driven content negotiation, that is by honouring the 
+                     request's Accept header and responding with the supported
+                     TD serialization and equivalent Content-Type header.
+                  </span>
+                  <span class="rfc2119-assertion" id="tdd-http-alternate-language">
+                     A Directory server MAY provide modified TDs or error responses using a different
+                     default language after server-driven content negotiation, 
+                     that is by honouring the request's Accept-Language header.
+                  </span>
 
                 <section id="exploration-directory-api-things" class="normative">
                     <h4>Things API</h4>
--- a/index.html
+++ b/index.html
@@ -1172,7 +1172,7 @@ img.wot-diagram {
                      default language after server-driven content negotiation, 
                      that is by honouring the request's Accept-Language header.
                   </span>
-                     The process of modifyig the default language of a TD using translations already
+                     The process of modifying the default language of a TD using translations already
                      provided in a TD is described in the WoT Thing Description 1.1 specification
                      [[wot-thing-description11]].
 
--- a/index.html
+++ b/index.html
@@ -1151,17 +1151,22 @@ img.wot-diagram {
                 </p>
 
                 <p>
+<!--
                   <span class="rfc2119-assertion" id="tdd-http-alternate-content">
                      A Directory server MAY provide alternative representations through
                      server-driven content negotiation, that is by honouring the 
                      request's Accept header and responding with the supported
                      TD serialization and equivalent Content-Type header.
                   </span>
+-->
                   <span class="rfc2119-assertion" id="tdd-http-alternate-language">
                      A Directory server MAY provide modified TDs or error responses using a different
                      default language after server-driven content negotiation, 
                      that is by honouring the request's Accept-Language header.
                   </span>
+                     The process of modifyig the default language of a TD using translations already
+                     provided in a TD is described in the WoT Thing Description 1.1 specification
+                     [[wot-thing-description11]].
 
                 <section id="exploration-directory-api-things" class="normative">
                     <h4>Things API</h4>
--- a/index.html
+++ b/index.html
@@ -1072,6 +1072,14 @@ img.wot-diagram {
                     </span>
                     This enables both machines and humans to know the high-level error class
                     and fine-grained details.
+                    <span class="rfc2119-assertion" id="tdd-http-error-response-utf-8">
+                        All HTTP API error responses described using Problem Details MUST be encoded using UTF-8.
+                    </span>
+                    <span class="rfc2119-assertion" id="tdd-http-error-response-lang">
+                        HTTP API error responses MAY report details in different languages using
+                        proactive negotiation, if the <code>Accept-Language</code> header field has been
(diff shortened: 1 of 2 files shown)


farshidtz:
> * I also noticed there was an assertion for this in self-description but not for directories, so while I was at it I added a similar assertion for directories

It is under Things API section:
> Directories MAY accept additional representations based on request's indicated Content-Type or Content-Encoding, and provide additional representations through server-driven content negotiation. This is useful for applications that require serializations other than JSON.

index.html
```diff
@@ -1145,6 +1150,19 @@ <h3>Directory Service API</h3>
...
+                     A Directory server MAY provide alternative representations through
+                     server-driven content negotiation, that is by honouring the 
+                     request's Accept header and responding with the supported
+                     TD serialization and equivalent Content-Type header.
+                  </span>
+                  <span class="rfc2119-assertion" id="tdd-http-alternate-language">
+                     A Directory server MAY provide modified TDs or error responses using a different
+                     default language after server-driven content negotiation, 
+                     that is by honouring the request's Accept-Language header.
+                  </span>
```
farshidtz, line 1164: We have an assertion bellow to allow directories to both receive and respond different content types. Should the directory also accept different languages from the client (when Content-Language is set on POST/PUT/PATCH).

https://github.com/w3c/wot-discovery/pull/318/files#diff-0eb547304658805aad788d320f10bf1f292797b5e6d745a3bf617584da017051R1210

Should these assertions be for the whole directory incl. Search and notification or just Things?

publication/3-cr/index.html
```diff
@@ -1624,13 +1632,22 @@ <h4>Listing</h4>
                                         and `sort_order` to choose the order
                                         (i.e. `asc` or `desc` for ascending and descending ordering).
                                     </span>
+                                    <span class="rfc2119-assertion" id="tdd-things-list-pagination-orderable">
+                                        A server MUST reject requests to sort on fields that do not have 
+                                        values that are orderable basic types, 
+                                        with a 501 (Not Implemented) status.
+                                    </span>
```
farshidtz, line 1639: I think this should be 400 (Bad Request). Not implemented is suitable when the server hasn't implemented a feature by choice.

index.html
```diff
@@ -1145,6 +1150,19 @@ <h3>Directory Service API</h3>
...
+
```
farshidtz, line 1165: We have an assertion bellow to allow directories to both receive and respond different content types. Should the directory also accept different languages from the client (when Content-Language is set on POST/PUT/PATCH).

https://github.com/w3c/wot-discovery/pull/318/files#diff-0eb547304658805aad788d320f10bf1f292797b5e6d745a3bf617584da017051R1210

I think these assertions should not be for the whole directory incl. Search and notification, but just for Things API.

index.html
```diff
@@ -827,11 +827,16 @@ <h2>Self-description</h2>
...
-                            The server MAY provide alternative representations through
+                            The self-description server MAY provide alternative representations through
                             server-driven content negotiation, that is by honouring the 
                             request's Accept header and responding with the supported
                             TD serialization and equivalent Content-Type header.
                         </span>
+                        <span class="rfc2119-assertion" id="self-http-alternate-language">
+                            The self-description server MAY provide modified TDs or error responses 
+                            using a different default language after server-driven content negotiation, 
+                            that is by honouring the request's Accept-Language header.
```
farshidtz, line 838: Is this the responsibility of the server to translate the TD? Or the translated field must already be in the TD?

farshidtz (changes requested):
Thanks. Looks good overal but some assertions need to be moved around and modified.

index.html
mmccool: No, what this means is that if "titles" is provided in a TD, for example, then if a particular language is requested, AND it is in the TD, then the returned TD would pull out the matching string and put it in "title" and change the default language.  There are assertions in the TD spec that say the set of languages provided must match in all instances of titles and descriptions (so if, for example, a German translation is provided for one it must be provided for all) and also includes a description of the process.  We don't have to repeat the definition but a references might be helpful.

publication/3-cr/index.html
mmccool: also... that was editing 3-cr/index.html for some reason!  Argh...

mmccool: fixed, moved (as are all other changes in this file...) to index.html

index.html
mmccool: I think we only have to worry about retrieval.  TDs themselves are multilanguage and include features to define their own default language.

index.html
mmccool: added a comment saying this process is defined in the TD spec

index.html
```diff
@@ -1145,6 +1158,24 @@ <h3>Directory Service API</h3>
...
+                     request's Accept header and responding with the supported
+                     TD serialization and equivalent Content-Type header.
+                  </span>
+-->
+                  <span class="rfc2119-assertion" id="tdd-http-alternate-language">
+                     A Directory server MAY provide modified TDs or error responses using a different
+                     default language after server-driven content negotiation, 
+                     that is by honouring the request's Accept-Language header.
+                  </span>
+                     The process of modifyig the default language of a TD using translations already
```
farshidtz, line 1175:
```suggestion
                     The process of modifying the default language of a TD using translations already
```

index.html
mmccool: resolution: won't do this

PR #323, opened 2022-05-27
Update Discovery overview figure and Architecture section

mmccool:
As mentioned in #299, the overview figure in the Architecture section is a little confusing, and when I looked into it I realized it also included some obsolete section references (the section numbering was changed after the figure was first created) and had a few other issues.  I redid it with the following changes, and also fixed a bunch of other things in this section.

Summary of changes to image:
- clearly differentiated introductions and explorations by position and color
- clearly indicated that only URLs get returned by introductions and get passed to intros
- ordered introductions by order of appearance in text
- removed section references (I considered just updating them, but I also want to fix a few other things that may cause further section renumbering, and don't want to create an annoying circular dependence)
- updated labels of each mechanism so they exactly match the section names
- added a few details, for example what fields to look for in DID, what rt to search for in CoRE, etc.
- tried to get rid of as many kinks in arrows as possible
- labelled phases (Introduction, Exploration, etc.)
- redrew the figure using inkscape (a native SVG editor) because the SVG exported by drawio had numerous problems especially with text; deleted drawio version; now the SVG and PNG versions are consistent

Other changes:
- Moved some of the text before the figure.  Before there was a single orphaned line right before the figure which kind of looked like a caption, but wasn't, and was easy to overlook.
- Inserted some paragraph breaks
- Tried to better explain the overall architecture, especially as regards to security and multiple results

Note: `images/overview-svc.{svg,png}` is not really part of this PR, it's not linked from anything.  It really should have been part of a follow-up PR.  See Issue #324

***
<a href="https://pr-preview.s3.amazonaws.com/mmccool/wot-discovery/pull/323.html" title="Last updated on May 30, 2022, 3:10 PM UTC (44c44a2)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-discovery/323/4f5ffec...mmccool:44c44a2.html" title="Last updated on May 30, 2022, 3:10 PM UTC (44c44a2)">Diff</a>


index.html
```diff
@@ -256,53 +256,112 @@ <h1>Terminology</h1>
...
+	       The intention is that Introductions are relatively open "first contact"
+         mechanisms to provide a starting point for the rest of the Discovery process.
+         In this document we specify details on several Introduction
+         mechanisms, suitable for different use cases, including both local and non-local
+         scenarios, but Introductions can in fact be provided any mechanism that can return a URL.
+	       Introductions, however, do not include any security or privacy controls and
+	       so should not provide metadata directly. 
+         Instead, the URLs provided by Introductions reference
+	       "Exploration" services.  Exploration services actually do provide metadata,
+         but only after suitable authentication and access controls have been applied.
```
farshidtz, line 293: Hard to follow. I suggest not shortening "Introduction mechanisms" to "Introductions"
```suggestion
	       The intention is that Introduction mechanisms are relatively open "first contact"
         mechanisms to provide a starting point for the rest of the Discovery process.
         In this document we specify details on several Introduction
         mechanisms, suitable for different use cases, including both local and non-local
         scenarios, but Introductions can in fact be provided any mechanism that can return a URL.
	       Introductions, however, do not include any security or privacy controls and
	       so should not provide metadata directly. 
         Instead, the URLs provided by Introduction mechanisms reference
	       "Exploration" services.  Exploration services actually do provide metadata,
         but only after suitable authentication and access controls have been applied.
```

index.html
```diff
@@ -256,53 +256,112 @@ <h1>Terminology</h1>
...
+	       can also be a <em>set</em> of Thing Descriptions.
+        </p>
+        <p>
+	       Each URL provided by the Introduction phase always points at an 
+         Exploration service endpoint that will return a single Thing Description.
+	       In the simplest case this is the Thing Description of a Thing 
+	       describing an IoT endpoint device.  
+         In the case of self-describing Things an Introduction URL might
+	       even point directly at an endpoint provided by a 
+         Thing serving its own Thing Description.
```
farshidtz, line 309: Two sentences telling the same thing?
```suggestion
         In the case of self-describing Things an Introduction URL might
	       even point directly at an endpoint provided by a 
         Thing serving its own Thing Description.
```

index.html
```diff
@@ -256,53 +256,112 @@ <h1>Terminology</h1>
...
+            Thing Description Directories that maintain large numbers of 
+            Thing Descriptions may also support queries that can be used to 
+            selectively retrieve Thing Descriptions.
+	        </li>
+	        <li>
+            The second special case is a 
+            <a href="#exploration-td-type-thinglink">Thing Link</a>.  
+            A Thing Link is also a Thing Description,
+            but rather than describing a Thing directly holds a link to
+            a Thing Description hosted elsewhere.
```
farshidtz, line 348:
```suggestion
            A Thing Link is also a Thing Description,
            but rather than describing a Thing directly, it holds a link to
            a Thing Description hosted elsewhere.
```

index.html
```diff
@@ -256,53 +256,112 @@ <h1>Terminology</h1>
...
+        <p>
+         It is <em>not</em> mandatory for the Discovery process to retrieve
+         the contents of Thing Description Directories and return them as 
+         part of the results, 
+         because in general this might result in a huge set of results.  
+         Instead the application
+         should scan the results for Thing Description Directory TDs and
+         decide whether to retrieve TDs from them, possibly selectively.
+         Likewise, it is not required to follow Thing Links automatically
+         instead the application may choose to follow them selectively.
```
farshidtz, line 363:
```suggestion
         Likewise, it is not required to follow Thing Links automatically;
         instead the application may choose to follow them selectively.
```

farshidtz (changes requested):
Added some comments inline.

About the new diagram, It looks simpler than before, but it is hard to follow:
- There are two kinds of squares with sharp and curly corners, not mapping consistently to the same kind of entity. 
- Arrows show HTTP requests on the left side, but something else on the right. Maybe the TD/TDs should be labels instead.
- In the end, the TDs points to an Application (Discoverer). Isn't the Discoverer the component that has the URLs and makes the GET request to begin with?
- The ThingLink arrows have no direction. Can be red from left to right or vice versa.

mmccool:
still some editorial changes, and some general improvements still needed in the figure, but I would like to merge (so it can be used for review) and go from there.

PR #325, opened 2022-05-27
Service exploration

### LINKED ISSUE #324: Allow "Service" Exploration Mechanism
While preparing PR #323 I realized that technically we can't have Introduction URLs pointing at "anything that can return a TD" as stated in the Discoverer section; we only have Self-Description and Directory.  Technically an Introduction URL can't just point at a web site hosting a TD if that TD describes *another* Thing, as opposed to itself.  We could use a Directory (that is also self-descriptive) but this is annoyingly heavyweight.

To resolve this we could add a third exploration mechanism, which I am tentatively proposing to call "Service", which supports the "any URL that can be used to fetch a TD".  Or rather "any URL that can be used to fetch a TD while providing suitable authentication and access controls".  It would be pretty much a clone of self-Description but would allow TDs that describe other Things.

mmccool:
Resolves #324 

Adds a new exploration mechanism which is basically the same as "self-description", but allows hosting of TDs that can describe Things other than itself, and allows hosting a TD on something which is not a Thing (e.g. a plain old web service), although there are some requirements for support of HEAD, etc. that need to be observed.

NOTE: Security assertions still need to be cleaned up.  For now I kept them (reworded a bit to be consistent with hosting) but added an ed note that the current ones are redundant/inconsistent and we should just refer to the other work we've done to specify these more carefully.

See also PR #330; *only one of #330 and this PR should be merged, they are two alternatives for solving the same problem*.

***
<a href="https://pr-preview.s3.amazonaws.com/mmccool/wot-discovery/pull/325.html" title="Last updated on Jun 1, 2022, 4:14 PM UTC (177d3e9)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-discovery/325/412212d...mmccool:177d3e9.html" title="Last updated on Jun 1, 2022, 4:14 PM UTC (177d3e9)">Diff</a>


k-toumura (approved):

index.html
```diff
@@ -788,6 +849,99 @@ <h4>`ThingLink`</h4>
...
+             However, a service distributing a TD 
+             using the following protocols 
+             is subject to the following constraints:
+            </p> 
+            <dl>
+                <dt>HTTP</dt>
+                <dd>
+                        <p class="ednote" title="Redundant">
+                        The following two assertions on TLS and authentication
+                        are redundant and not as specfic as those provided elsewhere.
```
farshidtz, line 873:
```suggestion
                        are redundant and not as specific as those provided elsewhere.
```

index.html
```diff
@@ -788,6 +849,99 @@ <h4>`ThingLink`</h4>
...
+                        Suggest replacing these here (and in all exploration sections) with appropriate
+                        references to security and privacy assertions given 
+                        in [[wot-architecture11]] and [[wot-thing-description11]]
+                        and in the Security Considerations and Privacy Considerations sections of this document.
+                        </p>
+                    <p>
+                        <span class="rfc2119-assertion" id="service-http-secure"> 
+                            An HTTP server providing a <a>TD</a> 
+                            SHOULD be using TLS.
```
farshidtz, line 882: TLS for?
```suggestion
                            An HTTP server providing a <a>TD</a> SHOULD use TLS when serving requests.
```

farshidtz (changes requested):

mmccool:
I cleaned up the PR to rebase it against main, so hopefully diffs make more sense now.  I also updated the figure and the text to use "Thing Description Service" instead of just "Service" ("Server" would have required a lot more changes...).  I also update the section header and the figure to use "Thing Description Directory" rather than just "Directory" to be consistent.

While I was at it I added some arrowheads to the optional backlinks in the figure.

index.html
mmccool: I will accept, although this whole assertion will be going away; I want to replace it with a reference to assertions we already have under Security and Privacy considerations.

mmccool:
I also moved the `overview-svc.{png,svg}` images to `overview.{png,svg}` since if this is merged, the new figure will not be an alternative, but the actual figure.

index.html
```diff
@@ -156,9 +156,10 @@
            leverages existing discovery mechanisms but does not directly
            expose metadata; they are simply used to discover Exploration
            services, which provide metadata but only after secure authentication
-           and authorization.  This document normatively defines two Exploration
-           services, one for WoT Thing self-description with a single 
-           WoT Thing Description and a searchable WoT Thing Description Directory
+           and authorization.  This document normatively defines three Exploration
+           services: for distributing a single WoT Thing Description from a regular
+           web service; for WoT Thing self-description providing a single 
+           WoT Thing Description; and a searchable WoT Thing Description Directory
```
farshidtz, line 162:
```suggestion
           and authorization.  This document normatively defines three Exploration
           services: for distributing a single WoT Thing Description from a regular
           web server; for WoT Thing self-description providing a single 
           WoT Thing Description; and a searchable WoT Thing Description Directory
```
Or "as a regular web service". I will use web server in the next suggestions.

index.html
```diff
@@ -223,8 +224,9 @@ <h1>Introduction</h1>
            WoT Thing Description Directory (WoT TDD) service that provides a search mechanism for
            collections of WoT Thing Descriptions that can be dynamically registered by Things or
            entities acting on their behalf.  The WoT Discovery mechanism however also supports
-           self-description by individual Things and one issue we address is how to distinguish
-           between these two approaches.
+           self-description by individual Things and distribution of single TDs from regular
+           web services.  One issue we address is how to distinguish
+           between these approaches.
```
farshidtz, line 229:
```suggestion
           self-description by individual Things and distribution of single TDs from regular
           web servers.  One issue we address is how to distinguish
           between these approaches.
```

index.html
```diff
@@ -286,7 +288,7 @@ <h1>Architecture</h1>
          mechanisms to provide a starting point for the rest of the Discovery process.
          In this document we specify details on several Introduction
          mechanisms, suitable for different use cases, including both local and non-local
-         scenarios, but Introductions can in fact be provided any mechanism that can return a URL.
+         scenarios, but Introductions can in fact be provided by any mechanism that can return a URL.
```
farshidtz, line 291:
```suggestion
         scenarios, but Introduction mechanisms can in fact be provided by any mechanism that can return a URL.
```

index.html
```diff
@@ -303,6 +305,8 @@ <h1>Architecture</h1>
         <p>
 	       Each URL provided by the Introduction phase always points at an 
          Exploration service endpoint that will return a single Thing Description.
+	       In the simplest case this is the Thing Description of a Thing 
+	       describing an IoT endpoint device.
```
farshidtz, line 309:
```suggestion
	       In the simplest case this is the Thing Description of a Thing 
	       describing an IoT device.  
```

index.html
```diff
@@ -984,9 +988,101 @@ <h3>Security Bootstrapping</h3>
...
             </section>
         </section>
+
+        <section id="exploration-service" class="normative">
+            <h2>Thing Description Service</h2>
+            <p>
+             Any web service that can be referenced by a URL and return a TD with
+             appropriate authentication and access controls can be used as an
+             exploration mechanism.  
+             This service does not need to be a Thing.
```
farshidtz, line 998:
```suggestion
            <h2>Thing Description Server</h2>
            <p>
             Any web service that can return a TD with
             appropriate authentication and access controls can be used as an
             exploration mechanism.  
             The server providing this service does not need to be a Thing.
```

index.html
```diff
@@ -984,9 +988,101 @@ <h3>Security Bootstrapping</h3>
...
+             In particular, a TD can be hosted on an ordinary web server
+             and referenced by its URL.  
+            </p>
+            <p>
+             However, a service distributing a TD 
+             using the following protocols 
+             is subject to the following constraints:
```
farshidtz, line 1005:
```suggestion
             However, a server distributing a TD 
             using the following protocols 
             is subject to the following constraints:
```

farshidtz (changes requested):
Thanks for the updates.
I think there is a confusing use of web server and web service. In particular a web server is the component that provides a web service. The service is providing a TD. 

https://stackoverflow.com/questions/2205788/is-there-a-difference-between-a-web-server-and-a-web-service

Unfortunately that means the diagrams should be updated too. IMO, the name of component should be "TD Server".



Why do we need the green boundary with a different label around the exploration services? I'd just remove those green dotted lines and labels.

index.html
mmccool: I did change "Thing Description Service" to "Thing Description Server" generally, but may have missed a few...

index.html
mmccool: outdated now

mmccool:
Also modified the figure to take out CoAP and multicast from CoRE-RD.  Technically you can use HTTP and unicast for CoRE-RD.  Also updated the class diagram.

Finally, a created another PR (PR #330, as a branch off this one) that merges self-description and TD server into one exploration mechanism, since a self-describing Thing is just a Thing that is also a TD server.  Unfortunately I branched that before merging Farshid's suggestions so if we go with PR #330 instead of this one (merging both does not make sense, they are alternative ways to address the same issue...) then we should do similar cleanups in the result.

BTW in either case we need to add "Thing Description Server" (short form, TD Server) to the Terminology definitions in Architecture.

mmccool:
merged alternative

PR #326, opened 2022-05-27
Self-Discovery Cleanup

mmccool:
Some cleanup to parallel the language in PR #325, add an ed note about the security assertions, and make sure the intro is actually narrowly talking about self-description (before it was pretty vague, and included mechanisms for distributing TDs that may not even use a URL...).

Note: I would like to merge this even without merging PR #325.  There is no dependency.  

However we probably should discuss WHY the intro to self-description was so broad and included not-URL mechanisms for distribution of TDs.  I seem to recall this had something to do with the use of files in the scripting API, etc. but I also feel that relative file URLs could be used to fill that need.

***
<a href="https://pr-preview.s3.amazonaws.com/mmccool/wot-discovery/pull/326.html" title="Last updated on May 30, 2022, 3:18 PM UTC (c46ab31)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-discovery/326/a47c788...mmccool:c46ab31.html" title="Last updated on May 30, 2022, 3:18 PM UTC (c46ab31)">Diff</a>


index.html
```diff
@@ -791,56 +791,66 @@ <h4>`ThingLink`</h4>
...
+                        Suggest replacing these here (and in all exploration sections) with appropriate
+                        references to security and privacy assertions given 
+                        in [[wot-architecture11]] and [[wot-thing-description11]]
+                        and in the Security Considerations and Privacy Considerations sections of this document.
+                        </p>
                     <p>
                         <span class="rfc2119-assertion" id="self-http-secure"> 
-                            The HTTP-based self-description SHOULD be over HTTPS (HTTP Over TLS).
+                            An HTTP-based self-description server
+                            SHOULD be using TLS.
```
farshidtz, line 817: Using TLS for what?
```suggestion
                            An HTTP-based self-description server SHOULD serve the requests over TLS.
```

farshidtz (changes requested):

PR #329, opened 2022-06-01
Content types: optional tree prefixes and parameters, split req/res content negotiation

### LINKED ISSUE #315: Return charset parameter in content-type
*Related to #281 but I decided to have a dedicated issue since it might be easier to track and the text in the spec is not clear*

I was implementing [Self-description](https://w3c.github.io/wot-discovery/#exploration-self) when one of my tests failed due to a wrong content type. Debugging I noticed that the web framework that I used automatically adds the charset parameter to the  `Content-Type header`. In practise, the value sent on the network was `application/td+json; charset=utf-8`. 

In understanding this behavior, I found [this](https://www.w3.org/International/articles/http-charset/index). Where it is stated that by default HTTP document charset is ISO-8859-1 and that's why is recommended to always explicitly state the charset parameter (browsers fall back to browsers use the reader's preferred encoding if there's none). 

Even though I'm aware that the default encoding of TD is UTF-8 I would change the text to prescribe `application/td+json; charset=utf-8` as Content-Type for consistency.

farshidtz:
- Allow content type with without `td`/`ld` tree prefixes (related to #299). This effectively allows using `application/json` instead of `application/td+json` and `application/ld+json`.
- Clarify content negotiation: split request/response assertions. Original assertion was hard to follow and missed Content-Encoding.
- Add assertion to explicitly mention the default JSON content type for self-description (related to #299). We already had `application/td+json`/`application/ld+json` as mandatory Content Type, so this isn't a technically a new requirement, but rather to clarify the spec.
- Allow parameters such as charset in content types (closes #315). This doesn't explicitly mention parameters or the charset, but instead makes the assertions less strict to allow such additions. As mentioned in https://github.com/w3c/wot-discovery/issues/315#issuecomment-1143320639, all `application/json` content types are already UTF-8 by specification. This PR makes it possible to explicitly specify that the charset is UTF-8 or something else.

***
<a href="https://pr-preview.s3.amazonaws.com/farshidtz/wot-discovery/pull/329.html" title="Last updated on Jun 13, 2022, 1:33 PM UTC (9319114)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-discovery/329/0549320...farshidtz:9319114.html" title="Last updated on Jun 13, 2022, 1:33 PM UTC (9319114)">Diff</a>


mmccool:
I see you updated the rules for this in TD Server for HTTP.  If we accept PR #334, we need equivalent updates for CoAP's Content-Format.

index.html
```diff
@@ -1064,18 +1064,21 @@ <h2>Thing Description Server</h2>
...
                         </span>
                         <span class="rfc2119-assertion" id="service-http-resp">
-                            A successful response from
-                            an HTTP-based TD Server providing a <a>TD</a> 
-                            MUST have a 200 (OK) status, contain a `application/td+json`
-                            Content-Type header, and the <a>TD</a> in the body.
-                        </span>
+                            A successful response from an HTTP-based TD Server providing a <a>TD</a>
+                            MUST have 200 (OK) status and the <a>TD</a> in the body.</span>
+                        <span class="rfc2119-assertion" id="service-http-resp-content-type">
```
mmccool, line 1069: I agree that breaking these into multiple assertions is a good idea.

index.html
```diff
@@ -1064,18 +1064,21 @@ <h2>Thing Description Server</h2>
...
+                            A successful response with JSON serialization MUST contain `application/json`
```
mmccool, line 1070: The use of "contain" is a bit strange here.  I can understand it if I interpret the content-types and defining sets, and td+json is a subset of json, but perhaps there is a better word than "contain" to make this clearer.

mmccool: Should we instead be more explicit and say "MAY" and "SHOULD" for the various options?

mmccool (changes requested):
I think we need a little more discussion, especially around the wording of the assertion with "contains" for application/json.   I think it is technically correct, but a little confusing.

relu91 (approved):
Looks good, regarding #315! thanks. About the content-type set to application/json I would probably need some clarification today. I thought Ben's feedback was about allowing clients to request application/json in the Accept header not giving the possibility to arbitrary return application/json or application/td+json. 

IMHO this assertion should already cover the WebThings gateway requirements:
> An HTTP-based TD Server providing a [TD](https://pr-preview.s3.amazonaws.com/farshidtz/wot-discovery/pull/329.html#dfn-td) MAY provide alternative representations through server-driven content negotiation, that is by honoring the request's Accept and Accept-Encoding headers and responding with the supported TD serialization and equivalent Content-Type and Content-Encoding headers

Isn't?

farshidtz:
> Looks good, regarding #315! thanks. About the content-type set to application/json I would probably need some clarification today. I thought Ben's feedback was about allowing clients to request application/json in the Accept header not giving the possibility to arbitrary return application/json or application/td+json.
> 
> IMHO this assertion should already cover the WebThings gateway requirements:
> 
> > An HTTP-based TD Server providing a [TD](https://pr-preview.s3.amazonaws.com/farshidtz/wot-discovery/pull/329.html#dfn-td) MAY provide alternative representations through server-driven content negotiation, that is by honoring the request's Accept and Accept-Encoding headers and responding with the supported TD serialization and equivalent Content-Type and Content-Encoding headers
> 
> Isn't?

This PR is addressing another feedback from Ben:
> I note that the response for this operation uses the application/ld+json content-type rather than application/td+json because the resource it returns is not a valid Thing Description, it's a list of Thing Descriptions. Could application/json be acceptable here too, as is used in the Search API?

farshidtz:
Issue for follow-up work: #341

mmccool:
propose merging, follow up with fixes in #341

PR #330, opened 2022-06-01
Self service exploration

mmccool:
Resolves Issue #324 

An alternative to PR #325 that makes self-description be a special case of a Thing Description Server (they have the same requirements, so... a self-describing Thing is just a Thing hosting a TD Server).  This makes the spec shorter and simpler and also makes testing easier (although some assertions starting with "self-" do get renamed).


index.html
```diff
@@ -717,16 +728,32 @@ <h2>Overview</h2>
                     alt="Exploration mechanisms high-level class diagram" />
                 <figcaption>
                     The high-level class diagram of the exploration mechanisms,
-                    depicting how Things expose <a>TDs</a>.
+                    depicting how Thing Description Servers and Thing Description Directories provide <a>TDs</a>.
+                    A Self-describing Thing is a special case of a Thing Description Server that is also a Thing.
+                    A Thing Description Directory hosts a Thing Description Server for each 
+                    Thing Description it contains.
```
farshidtz, line 734: It doesn't really host a server for each TD. Literally doing so will be very inefficient.
```suggestion
                    A Thing Description Directory acts as a Thing Description Server for each 
                    Thing Description it contains.
```

farshidtz (changes requested):
Thing Description Server (green) seems wrong. I think this is meant as self-describing thing?


Why is TDD hosting zero or more TD servers? The TDD only need one server to serve all the TDs it contains. The self-describing TDD inherits TDD and self (+TD server) and as a result can provide its own TD and contain others.

index.html
mmccool: need a better word than "server", is really an endpoint

PR #334, opened 2022-06-13
Add text for discovery using CoAP

JKRhb:
This PR adds a number of CoAP-specific assertions for discovery from Thing Description Servers to the document, modelled after the ones for HTTP.

***
<a href="https://pr-preview.s3.amazonaws.com/JKRhb/wot-discovery/pull/334.html" title="Last updated on Jun 13, 2022, 2:00 PM UTC (35d7b3c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-discovery/334/0549320...JKRhb:35d7b3c.html" title="Last updated on Jun 13, 2022, 2:00 PM UTC (35d7b3c)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -2980,6 +2980,12 @@ img.wot-diagram {
                 The HTTP/1.1 chunked transfer encoding is not supported in HTTP/2.
                 <!-- </span> -->
                 </li>
+                <li>
+        <!-- <span class="rfc2119-assertion" id="perf-inc-chunked-3"> -->
+                CoAP servers should handle the data incrementally using
+                block-wise transfer [[RFC7959]].
+               <!-- </span> -->
+                </li>
             </ul>
             <p> 
             Most HTTP servers and clients automatically process the data that is transferred 
--- a/index.html
+++ b/index.html
@@ -1112,6 +1112,71 @@ img.wot-diagram {
                     </p>
                 </dd>
             </dl>
+            <dl>
+                <dt>CoAP</dt>
+                <dd>
+                <p>
+                    <span class="rfc2119-assertion" id="service-coap-secure">
+                        A CoAP-based TD Server providing a <a>TD</a> SHOULD use the appropriate transport
+                        security mechanism for the transport protocol used (i.e., DTLS for CoAP over UDP
+                        or TLS for CoAP over TCP).
+                    </span>
+                    <span class="rfc2119-assertion" id="service-coap-access-control">
+                        A CoAP-based TD Server providing a <a>TD</a>
+                        SHOULD provide the resource only after performing necessary
+                        authentication and authorization.
+                    </span>
+                </p>
+                <p>
+                    <span class="rfc2119-assertion" id="service-coap-method">
+                        A CoAP-based TD Server providing a <a>TD</a>
+                        MUST serve that resource with a `GET` method.
+                    </span>
+                    <span class="rfc2119-assertion" id="service-coap-resp">
+                        A successful response from a CoAP-based TD Server providing a <a>TD</a>
+                        MUST have a 2.05 (Content) status, contain a Content-Format option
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -1112,6 +1112,71 @@ <h2>Thing Description Server</h2>
...
                 </dd>
             </dl>
+            <dl>
+                <dt>CoAP</dt>
+                <dd>
+                <p>
+                    <span class="rfc2119-assertion" id="service-coap-secure">
+                        A CoAP-based TD Server providing a <a>TD</a> SHOULD use the appropriate transport
+                        security mechanism for the transport protocol used (i.e., DTLS for CoAP over UDP
+                        or TLS for CoAP over TCP).
```
JKRhb, line 1122: I wasn't sure if OSCORE should be mentioned here as an alternative/additional method for securing CoAP messages (CC @j1y3p4rk).

index.html
```diff
@@ -1112,6 +1112,71 @@ <h2>Thing Description Server</h2>
...
+                </p>
+                <p>
+                    <span class="rfc2119-assertion" id="service-coap-method">
+                        A CoAP-based TD Server providing a <a>TD</a>
+                        MUST serve that resource with a `GET` method.
+                    </span>
+                    <span class="rfc2119-assertion" id="service-coap-resp">
+                        A successful response from a CoAP-based TD Server providing a <a>TD</a>
+                        MUST have a 2.05 (Content) status, contain a Content-Format option
+                        with value 432 (`application/td+json`), and the <a>TD</a> in the payload.
```
mmccool, line 1138: we can leave this as is, but I'm wondering if the part here about the Content-Format (and the corresponding part under the HTTP section) conflicts with what we have done for Directories to make this more flexible.  HOWEVER we can go ahead and merge this and maybe just raise an issue to think about making these consistent.

index.html
```diff
@@ -1112,6 +1112,71 @@ <h2>Thing Description Server</h2>
...
+                        server-driven content negotiation, that is by honouring the
+                        request's Accept option and responding with the supported
+                        TD serialization and equivalent Content-Format option.
+                    </span>
+                </p>
+                <p>
+                    <span class="rfc2119-assertion" id="service-coap-size2">
+                        A CoAP-based TD Server providing a <a>TD</a>
+                        SHOULD respond to requests containing a Size2 option by including
+                        the size estimate of the TD in its next response.
```
mmccool, line 1154: This also reminds me that for CoRE-RD and CoRE-Link format, we may want to make the "sz" parameter mandatory.  Not an issue for this PR, however...

index.html
```diff
@@ -2915,6 +2980,12 @@ <h2>Incremental Transfer</h2>
                 The HTTP/1.1 chunked transfer encoding is not supported in HTTP/2.
                 <!-- </span> -->
                 </li>
+                <li>
+        <!-- <span class="rfc2119-assertion" id="perf-inc-chunked-3"> -->
```
mmccool, line 2984: we don't currently define CoAP-based directories (these are purely HTTP) put putting this here but commented out is helpful in case we ever do want to extend directories

relu91 (approved):
PR looks good, talked with @JKRhb offline, some discussion points for today's call:
- The counterpart operation for knowing TD size in coap is a message with Size2 option. However, it is also part of the block-wise transfer mode of CoAp. Maybe it would be ok to just reccomand the support for block-wise transfer
- incremental transfer section contains assertions but they are commented out, is that on purpose?
- we should stress the fact that block-wise transfer must be supported for CoAp over udp.

index.html
JKRhb: That simplifies things a lot :) Removed the assertions in 35d7b3c5af7742a76336d3b58a6f82a296c5d4c3.

farshidtz:
Issue to track follow-up work: https://github.com/w3c/wot-discovery/issues/340

PR #337, opened 2022-06-13
Clean up Discoverer section

mmccool:
Remove some text from the "Discovery Process" section:
- Remove mention of file URLs.  It's not clear we want to support these, and if we do, they should be mentioned as a protocol in the TD Server section, and they are not.
- Remove examples around CoAP, and statements saying that other protocols than HTTP can be used.  This does not disallow CoAP or other protocols, as they were only examples, but this decouples this section from the question of whether we support CoAP or not.  So now the specification of what protocols are allowed are JUST in the relevant "exploration" sections.
- Got rid of an unnecessary editor's note.

***
<a href="https://pr-preview.s3.amazonaws.com/mmccool/wot-discovery/pull/337.html" title="Last updated on Jun 13, 2022, 1:36 PM UTC (6fe44ae)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-discovery/337/0549320...mmccool:6fe44ae.html" title="Last updated on Jun 13, 2022, 1:36 PM UTC (6fe44ae)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -380,27 +380,28 @@ img.wot-diagram {
 	and Thing Links and use specific affordances and links provided in them.  
 	Conversely, a Consumer may not support Discovery, although it is recommended
 	[[wot-architecture11]].
-        </p>
-	<p class="ednote" title="Consumers SHOULD support Discovery">
-	Note: the WoT Architecture specification only says that Consumers SHOULD support Discovery.  
-	</p>
-        <p>
+</p>
+<p>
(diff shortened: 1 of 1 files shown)


relu91 (approved):
I think the remove statements were nice examples/clarifications, but I agree that they are quite controversial given the current status. Ok for merge 👍🏻

JKRhb (approved):

PR #309, opened 2022-05-13
Draft Explainer

mmccool:
Ready for review.  Please check carefully for technical accuracy.  We should also decide [whether we should really allow non-HTTP self-description explorations given that our spec has <strike>almost</strike> no assertions constraining them.](https://github.com/w3c/wot-discovery/issues/314)  The explainer would have to be updated if we do decide to support non-HTTP explorations (which the current specification DOES allow).

PR #345, opened 2022-06-13
Explainer Fixes

### LINKED ISSUE #314: Self-Description HTTP only?
Currently the spec has a small parenthetical "e.g." for Self-Description implying CoAP (and possibly other) protocols may be used, but provides assertions only for HTTP.  Currently the explainer draft is written assuming we are only using HTTP.  To keep scope under control, I would like to remove the "(e.g. ...)" in the introduction to this section and normatively define only HTTP (for now; we can consider other protocols in another round).

mmccool:
Resolves #314

- Remove "RESTful" from various places since the Directory API is not technically RESTful (as it includes event notification)
- Add TD Server, with self-description as a special case
- Explicitly allow CoAP in addition to HTTP (at least for TD servers)
- Add link to WoT Discovery specification

### DIFF
--- a/explainer/Explainer.md
+++ b/explainer/Explainer.md
@@ -26,7 +26,7 @@ first-contact mechanisms suitable for different scales; and by providing optiona
 semantic search and self-description capabilities.
 
 There are still a number of limitation in the current specification, which
-primarily focuses on CoAP and HTTP and had to defer some important goals, such as
+primarily focuses on HTTP and CoAP, and had to defer some important goals, such as
 geolocation queries.  However, the specification attempts to address the above
 core issues and is intended to serve as a foundation for future work while
 enabling, in its current form, use cases supporting IoT integration
--- a/explainer/Explainer.md
+++ b/explainer/Explainer.md
@@ -1,5 +1,7 @@
 # Web of Things (WoT) Discovery Explainer
-The purpose of WoT Discovery is to define a mechanism to distribute WoT Thing Descriptions
+The purpose of 
+[WoT Discovery](https://w3c.github.io/wot-discovery/) 
+is to define a mechanism to distribute WoT Thing Descriptions
 in a manner that is accessible but also satisfies security and privacy objectives.
 It is an optional feature of the WoT Architecture in that WoT TDs can also be 
 managed and obtained by other means, for example by using a normal database.  However, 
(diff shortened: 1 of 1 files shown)


explainer/Explainer.md
```diff
@@ -187,18 +194,21 @@ use cases like a Smart City where there
 might be thousands of Things to select from.
 However, it can also be used in a Smart Home with only a few dozen Things.
 
-Self-description is another exploration mechanism that
-provides the ability of an HTTP-based Thing to
-return a TD directly.  There is also an Introduction mechanism that allows
+A TD Service is another exploration mechanism that allows a single
```
relu91, line 197:
```suggestion
A TD server is another exploration mechanism that allows a single
```
?

relu91 (approved):
Changes look good to me. Just one minor typo(?):

explainer/Explainer.md
mmccool: good catch, fixed.  But I used uppercase to be consistent with the spec.

explainer/Explainer.md
```diff
@@ -19,12 +21,12 @@ cannot be modified and so can only be described externally.
...
 The WoT Discovery specification attempts to meet these challenges by defining
 a multistage discovery process with authorization checkpoints to meet privacy and
-security goals; by using scalable RESTful API designs; by supporting a variety of
+security goals; by using scalable network API designs; by supporting a variety of
 first-contact mechanisms suitable for different scales; and by providing optional
 semantic search and self-description capabilities.
 
 There are still a number of limitation in the current specification, which
-primarily focuses on HTTP and had to defer some important goals, such as
+primarily focuses on CoAP and HTTP and had to defer some important goals, such as
```
JKRhb, line 29: As the specification currently focuses on HTTP and in order to be consistent with the order in the rest of the document, I think CoAP and HTTP could be swapped here. Other than that, looks good to me :) Thank you, @mmccool!

```suggestion
primarily focuses on HTTP and CoAP, and had to defer some important goals, such as
```

farshidtz:
Another needed fix: I realized the change in https://github.com/w3c/wot-discovery/pull/309/commits/1f614705f0c1f582e7703d522003918a54efcacd was a mistake suggested by me in #309. According to the TM, the only required affordance is indeed `things` for listing.

PR #350, opened 2022-06-14
Remove reference to issue 86

### LINKED ISSUE #86: "Thing description" for Directory should be a "Thing Model"

mmccool:
Resolves #86 

There was an old reference to issue #86 which basically said "we should use a Thing Model".  Well, we ARE using a Thing Model now, so I closed issue 86 and removed the reference to it in the text.  I also added some missing markup in the intro paragraph and tweaked the opening sentence again to make it clearer (I hope) that the TM applies to Thing Description Directories as a class.

***
<a href="https://pr-preview.s3.amazonaws.com/mmccool/wot-discovery/pull/350.html" title="Last updated on Jun 20, 2022, 2:20 PM UTC (88d50f5)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-discovery/350/a3ab94c...mmccool:88d50f5.html" title="Last updated on Jun 20, 2022, 2:20 PM UTC (88d50f5)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -2379,26 +2379,14 @@ img.wot-diagram {
 
                 <section id="directory-api-spec">
                     <h4>API Specification (Thing Model)</h1>
-
-                    <p class="issue" data-number="86">
-                        To do: add a
-                        <a href="https://w3c.github.io/wot-thing-description/#thing-model">Thing Model</a>
-                        (`directory.tm.json`)
-                        that can derive to
-                        <a href="https://github.com/w3c/wot-discovery/blob/main/implementations/example-directory-oauth2.td.json">this TD</a>
-                        as an example.
-                        The <a>TM</a> should describe all interaction affordances and indicate which ones
-                        are required (i.e. mandatory by this spec). Moreover, it should use placeholders
-                        for deployment-specific attributes.
-                        Informative affordances (JSONPath and XPath) should be defined as such.
-                    </p>
-
-                    The API of the directory is specified as a <a>Thing Model</a>.
+                    <p>
+                    A template for the API of Thing Description Directories is given here
+                    as a <a>Thing Model</a>.
                     The Thing Model alone should not be considered as the reference
                     to implement or interact with a directory.
                     The full specification is available as human-readable text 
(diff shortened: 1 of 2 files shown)


index.html
```diff
@@ -2379,26 +2379,14 @@ <h4>Semantic search: SPARQL</h4>
...
-                        To do: add a
-                        <a href="https://w3c.github.io/wot-thing-description/#thing-model">Thing Model</a>
-                        (`directory.tm.json`)
-                        that can derive to
-                        <a href="https://github.com/w3c/wot-discovery/blob/main/implementations/example-directory-oauth2.td.json">this TD</a>
-                        as an example.
-                        The <a>TM</a> should describe all interaction affordances and indicate which ones
-                        are required (i.e. mandatory by this spec). Moreover, it should use placeholders
-                        for deployment-specific attributes.
-                        Informative affordances (JSONPath and XPath) should be defined as such.
```
farshidtz, line 2393: The issue isn't fully resolved. In particular, we still have the informative JSONPath and XPath in there.

I suggest we simply remove them since the TM is normative. The text already covers everything that is necessary for implementations. We could also move them to another TM but that's really unnecessary.

mmccool: We could also just add a comment to the TM, in the form of a "description", that the corresponding parts of the TM are not normative.

mmccool:
Updated.
- Text revised to make clearer that the TM IS normative, "except where noted", but that the rest of the specification in the section is applicable.
- Added notes in descriptions of searchJSONPath and searchXPath saying that "This affordance is not normative and is provided for information only."
- Added following sentence to end of intro text: "The searchJSONPath and searchXPath affordances given in this Thing Model are not normative and are provided for information only."

PR #357, opened 2022-06-24
Draft for publication (says is CR, will be changed to WD)

mmccool:
Working on creating static HTML files for release.  This updates the files in publication/3-cr.  @ashimura please let me know if these should go in a different place (e.g. 4-cr) but my understanding was we never published 3-cr.   I also based this release on the "main" branch so we should create a diff relative to the cr-candidate branch for the Resolution next week.

Current status: still two errors on PubRules checker, but one seems spurious (it is complaining about a missing wide review link when there is, in fact, the exact text it is recommending already in the document) and the second is about using media type text/plain (due to where I am serving it from; I may need to check it into a side branch on the wot-discovery repo so I can use github.io to serve it to the pubrules checker).  Link checker also refuses to check files served as text/plain... sigh.

The scanning process discovered two duplicate ids, which I resolved with PR #358:
* line 2057 column 29 - Warning: <span> anchor "tdd-things-retrieve-resp-content-type" already defined
* line 3456 column 17 - Warning: <span> anchor "priv-loc-gen-ids" already defined

There were also a number of "empty" span's and p's which were "trimmed", which I find suspicious.  This might be a case of a span around a p confusing the parser, but the "trimming" might also be removing assertion markup.  I am still investigating, so will hold this PR as draft for now.


mmccool:
Note: would like to merge, but this is NOT a CR transition, it will be a WD.  I will do another PR to update that.  But I'd like to merge this so I can run the Link Checker.

Issue #371, opened 2022-07-04
Fixes to 3-WD


mmccool:
* Farshid's affiliation should be removed, and changed to "formerly {F-G}, currently Invited Expert"

mmccool:
When I was putting together PR #372 I noticed a couple more things to fix: an extraneous ed note, PR #353 on tracking/profiling, and updating the change log.  So this PR does those too, and I added them to the description.

mmccool:
PR #373 addresses the point related to Issue #99.  I added checkboxes to the description so we can check off everything that we have a PR for now...

mmccool:
So current WD links to the implementation report in the repo, using github.io.  We should really make a static snapshot and update this URL to point at the snapshot.  For now I will update the implementation to fix the internal links and dates, and then close this issue, but for the final CR version we should point at a snapshot.

See also issue #391 - the link exists only in the WD, not the editor's draft.  In general there are a number of fixes that only made it into the WD and we should back-patch them into the main branch.

PR #373, opened 2022-07-09
Add JSON Schema of extensions, TD minimal validation reference

### LINKED ISSUE #99: Define requirements and add section for validation
How much validation does a directory need to do?
1. If it only supports syntactic queries, is JSONSchema validation sufficient?   Do we also use SHACL? 
2. If we support SPARQL, do we need to do more to make sure the ontologies used are correct?
3. Is there a subset of TD assertions we should check?

farshidtz:
Related to #371 
Closes #99

***
<a href="https://pr-preview.s3.amazonaws.com/farshidtz/w3c-wot-discovery/pull/373.html" title="Last updated on Jul 11, 2022, 2:34 PM UTC (280e35e)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-discovery/373/eb657af...farshidtz:280e35e.html" title="Last updated on Jul 11, 2022, 2:34 PM UTC (280e35e)">Diff</a>


index.html
```diff
@@ -2401,9 +2405,8 @@ <h4>API Specification (Thing Model)</h1>
                     given in this Thing Model
                     are not normative and are provided for information only.
                     </p>
-                    <pre class="advisement json">
-                        <div data-include='directory.tm.json'></div>
-                    </pre>
+                    <pre class="advisement json"
+                         data-include='directory.tm.json'></pre>
```
farshidtz, line 2409: This is change is to remove an extra line-break from the rendered output.

index.html
```diff
@@ -2033,16 +2033,20 @@ <h4>Listing</h4>
                         
                     </section>
 
-                    <section id="validation" class="normative">
+                    <section id="exploration-directory-api-things-validation" class="normative">
```
farshidtz, line 2036: Modified left-over generic ID from when validation section was outside of TDD's API.

index.html
mmccool: Note that I in fact inlined this in the 3-wd publication.

mmccool:
So in general looks good, but I think we should update the assertion to explicitly allow either TD 1.0 or TD 1.1 versions.  I am ok with merging if my suggested change is made.  Also, this PR is against the main branch, but I would like to update the 3-wd version (the resolution we made last week in the main call allows for that).  As part of doing that I will be inlining the data-include to make the file self-contained.

mmccool (approved):

PR #374, opened 2022-07-10
Remove extra quotation from id in manual.csv

farshidtz:


mmccool:
bug fix, no problem merging.  I only wonder how it happened in the first place...

mmccool (approved):

Issue #382, opened 2022-07-29
How should non-URN `id`s be encoded in `/things/{id}` URLs and `Location` headers?


benfrancis:
The Thing Description specification says that an id can be any URI. The examples in the Thing Discovery specification are all URNs, but that's only one type of URI.

From reading the specification, my assumption is that if you want to create a Thing Description with an HTTP URL as its ID, you'd have to submit an HTTP request like:

`PUT http://localhost:3000/things/https://mywebthingserver.com/things/lamp`

That looks a little strange, but it is a valid URL as far as I can tell. If you try that with [Zion](https://github.com/vaimee/zion) you get a 404 error:

```
{
    "statusCode": 404,
    "message": "Cannot PUT /things/https://mywebthingserver.com/things/lamp",
    "error": "Not Found"
}
```

Perhaps URLs as IDs like that should be URL encoded? E.g.

`PUT http://localhost:3000/things//https%3A%2F%2Fmywebthingserver.com%2Fthings%2Flamp`

Zion will accept that request, but you can end up with `http://localhost:3000/things/https%3A%2F%2Fmywebthingserver.com%2Fthings%2Flamp42` as the URL of the Thing Description and `https://mywebthingserver.com/things/lamp` as the `id` member inside the Thing Description. It actually doesn't seem to check that they are the same.

I've been wondering for a while how this would work in practice. Should the server accept `PUT /things/https://mywebthingserver.com/things/lamp`, or should it be URL encoded as `PUT /things/https%3A%2F%2Fmywebthingserver.com%2Fthings%2Flamp`?

If it's the former, then maybe an example would help illustrate that?

If it's the latter, then maybe some text is needed in the specification to deal with this?

benfrancis:
For the record, currently if you add a web thing with an HTTP URL as an ID to WebThings Gateway, it gets given a URL like:

`https://user.webthings.io/things/https---mywebthingserver.com-things-lamp` (with all the non-alphanumeric characters replaced with dashes)

Since WebThings Gateway acts as a gateway, not just a directory, it also generates a new Thing Description for that Thing, using that URL as the new `id` member.

relu91:
Good point, my take on this issue was that people would encode IDs that contain problematic characters, even though usually web devs are used to encode query parameters, not paths. for example I think postman autoenconding feature would break if you try to issue something like `http://localhost:3000/things/https://mywebthingserver.com/things/lamp`

>  then maybe some text is needed in the specification to deal with this?

+1

[...]

PR #383, opened 2022-07-31
Clarify Anonymous ID assignment and Location header

### LINKED ISSUE #379: Clarify `Location` header content
I'm following these assertions:
> Upon successful processing, the server MUST respond with 201 (Created) status and a Location header containing a system-generated identifier for the TD.

The text informally redirects the reader to the definition of this system-generated identifier.

> The local identifier SHOULD be a UUID Version 4, presented as a URN [[RFC4122](https://w3c.github.io/wot-discovery/#bib-rfc4122)].

In my understanding, the two would allow location headers that contain simply the UUID 4 of the generated TD. The assumption is also followed in the @farshidtz test bed, where the content of the Location Header is concatenated to the server `basepath` plus `/things`. This, however, goes against the [definition](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Location) where it is clearly mentioned that the content of the header must be an URL:
> A relative (to the request URL) or absolute URL.

Is this bug? Or I am missing something?

farshidtz:
Related to #379 (discussion Aug 1: does not completely address issue).   See also #382 

An assertion is being added but it doesn't require any technical changes for the implementations, since we already require returning the system-generated ID in the location header.

***
<a href="https://pr-preview.s3.amazonaws.com/farshidtz/w3c-wot-discovery/pull/383.html" title="Last updated on Aug 1, 2022, 2:32 PM UTC (4705a22)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-discovery/383/57cca9c...farshidtz:4705a22.html" title="Last updated on Aug 1, 2022, 2:32 PM UTC (4705a22)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1570,7 +1570,7 @@ img.wot-diagram {
                                     JSON serialization of TD.</span>
                                 The TD object is validated in accordance with [[[#exploration-directory-api-things-validation]]].
                                 <span class="rfc2119-assertion" id="tdd-things-create-anonymous-id">
-                                    The directory MUST assign a local identifier to the <a>Anonymous TD</a>
+                                    The directory MUST assign a local identifier to any <a>Anonymous TD</a>
                                     to enable local management and retrieval from the directory.</span>
                                 The scheme of the system-generated ID is described in [[[#exploration-directory-anonymous-td]]].
 
--- a/index.html
+++ b/index.html
@@ -1330,7 +1330,7 @@ img.wot-diagram {
                     to enable management and retrieval of such TDs from the directory.
                     <span class="rfc2119-assertion" id="tdd-anonymous-td-identifier">
                         In situations where the server exposes an Anonymous TD (e.g. retrieval, listing, search),
-                        it MUST add the local identifier to the TD to allow local referencing.</span>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -1569,10 +1569,16 @@ <h4>Creation</h4>
                                     The request SHOULD contain `application/td+json` Content-Type header for 
                                     JSON serialization of TD.</span>
                                 The TD object is validated in accordance with [[[#exploration-directory-api-things-validation]]].
+                                <span class="rfc2119-assertion" id="tdd-things-create-anonymous-id">
+                                    The directory MUST assign a local identifier to the <a>Anonymous TD</a>
```
mmccool, line 1573:
```suggestion
                                    The directory MUST assign a local identifier to any <a>Anonymous TD</a>
```

mmccool (approved):

mmccool:
Looks like this repo is also having IPR issues (I am having the same problem with my commits).

mmccool:
No problem in theory, but should resolve the IPR issue first.

mmccool:
Note that @farshidtz is now (once again) an official WoT participant, and this *is* a substantive contribution, and IPR review requirements have been satisfied.

PR #401, opened 2022-08-29
DID service names in document and fix Directory Description

### LINKED ISSUE #65: Create typed links in DID Documents for WoT Directories
These should be consistent with a general model that we should also use for similar things, i.e. CoRE-RD and DNS-SD.   We should discuss with DID the process for how to document a link type for use in DID Documents.

### LINKED ISSUE #393: Update DID Service Names
The DID service names are not fully specified in the document (only appearing in an example) and should be "WotThing" and "WotDirectory".   Updates to https://www.w3.org/TR/wot-discovery/#introduction-did-sec needed.

As this may involve normative changes we should do it before CR.  For consistency we may also want to add a section under "IANA Considerations" (which, BTW, is also misnamed since it has subsections that refer to several things already not managed by IANA, e.g. CoRE RD...)

It also seems these service names need to be "vocabulary" in the JSON-LD sense (a DID document is a JSON-LD document) and so defined in a context file (or in an `@context` section of the DID document?).  There are also potential problem with prefixes, etc.  So what we *actually* have to do is define URLs for these terms and then discovery needs to *resolve* those terms with those specific URLs.   I suggest putting these terms into the wot-discovery ontology so we can use the base URL already defined for that context (although this is not QUITE correct, since these terms are for DID Documents, not TDs; ideally we would define another context file and another base URL for it, which we would then have to register?)

mmccool:
Resolves #393 
Resolves #65 

- update assertion in introduction-did-sec to use WotThing and WotDirectory
- add an example (now one examples for each)
- comment out reference to issue 65 (this PR should resolve it)
- add DID section under IANA considerations (although... they do not use IANA to register names, so this is not QUITE right - so allocated an "Other Name Registration Consideration" section at the same level and right after the IANA Considerations section)
- Noticed various places still using "Directory Description" in Introductions so changed that to "Thing Description" where appropriate

***
<a href="https://pr-preview.s3.amazonaws.com/mmccool/wot-discovery/pull/401.html" title="Last updated on Aug 29, 2022, 2:34 PM UTC (520dbd1)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-discovery/401/73b7647...mmccool:520dbd1.html" title="Last updated on Aug 29, 2022, 2:34 PM UTC (520dbd1)">Diff</a>


mmccool:
Registration itself still needs to be done with DID, see https://github.com/w3c/wot-discovery/issues/402

PR #421, opened 2022-10-03
Oct 2022 Testfest Updates to Implementation Report

mmccool:
- assertion wording change (ref update: CoRE-RD spec is now RFC9176)


k-toumura:
[rendered version](https://cdn.statically.io/gh/w3c/wot-discovery/0c2a7ea481c0c628132588e6895ddd439f96f27c/testing/report.html)

relu91:
I spotted a couple of assertions we should already cover with zion: https://w3c.github.io/wot-discovery#tdd-http-unsupported-feature https://w3c.github.io/wot-discovery#tdd-things-create-anonymous-id. Perhaps our results are old?

mmccool:
Things are stable now but there were a few changes in the last month.  The current PR updating the Implementation Report for the testfest includes and updated template.csv file.  What we need is some tool that can check the list of assertions in two files and let us know if the lists are different, ignoring order and other columns.  I don't know of any CSV tool that does that but it would be relatively easy to whip up in Node.js using Map and csvtojson...

relu91:
> Things are stable now but there were a few changes in the last month. This PR updating the Implementation Report for the testfest also includes an updated template.csv file. What we need is some tool that can check the list of assertions in two files and let us know if the lists are different, ignoring order and other columns. I don't know of any CSV tool that does that but it would be relatively easy to whip up in Node.js using Map and csvtojson...

found this https://github.com/simonw/csv-diff, should do the work for us. I'm not seeing the updated `template.csv` in the [changed files](https://github.com/w3c/wot-discovery/pull/421/files). Should I contribute directly to this PR or in the testing repo?

mmccool:
@relu91 best to contribute to the events dir, I will copy things over to the data dir when I process the updates.

mmccool:
Said in meeting I would merge today, but would like to include some of the results from zion.  @relu91 if you can contribute zio results to the testing repo that would be great, then I can include them in the updates we can discuss for the Monday discovery call.

mmccool:
I see there are also some updates from logilab.  I'm still not liking the completeness of results, especially for security assertions.  Will discuss today in the call.

relu91:
Ok perfect, cause I'm trying to understand how to contribute to the results properly but it seems that https://github.com/farshidtz/wot-discovery-testing has not been updated in a while, therefore I'm not sure that I'm using the latest assertions list. Let's talk about it during the call.

mmccool:
To be as clear as possible...

* Latest full assertions list is [here](https://github.com/w3c/wot-discovery/blob/main/testing/template.csv) (this is automatically generated from the index.html whenever I update the implementation report)
* Latest *manual* assertion list is [here](https://github.com/w3c/wot-discovery/blob/main/testing/manual.csv).  This is manually maintained and should be revised whenever the set of automatically tested assertions changes, or there are new assertions that we do not have automatic tests for.
* New results for an implementation should be put [here](https://github.com/w3c/wot-testing/tree/main/data/input_2022/Discovery).  I will copy csv files out of the Results directory to wot-discovery when I update the reports; the directories for each implementation can have separate files for automatic and manual results but these should be combined for the version in Results.

mmccool:
above notes added to README here: https://github.com/w3c/wot-testing/blob/main/data/input_2022/Discovery/README.md

mmccool:
I also just updated https://github.com/w3c/wot-discovery/blob/main/testing/README.md and added a link to wot-testing.  Hopefully that will help somewhat with any confusion...

Issue #416, opened 2022-09-26
clean up logilab results


mmccool:
- seem to be duplicate result files
- reporting "not implemented" for discoverer assertions, but these should be "null"

wiresio:
See https://github.com/w3c/wot-discovery/pull/422

PR #422, opened 2022-10-10
Clean up Siemens/Logilab results

wiresio:
Addresses https://github.com/w3c/wot-discovery/issues/416


mmccool:
Can you also please commit this to wot-testing in data/input_2022?

PR #423, opened 2022-10-12
Add tdd-things-create-anonymous-id to manual tests

farshidtz:
This is a relatively new assertion which is mostly informational and not directly testable from outside the directory:
> The directory MUST assign a local identifier to any Anonymous TD to enable local management and retrieval from the directory.

There are other assertions that specifically prescribe the ID format and the response from the directory which are testable.


relu91 (approved):
Talked, offline with @farshidtz. We agreed that there are some aspects of that assertion that can be tested automatically (i.e. if the TDD answer correctly after the creation of an Anonymous TD, then the assertion is true), but for the time being, we should move it to manual.

PR #427, opened 2022-10-14
Update Oct 2022 impl report

mmccool:
- sync new results for VAIMEE/Zion, TinyIot, logilab, etc.
- make sure changes synced with wot-testing, using latest versions
- update implementation report, including changes to which assertions are tested manually


k-toumura:
[Rendered version (addfa05)](https://cdn.statically.io/gh/w3c/wot-discovery/addfa0561cdf58ed43a719f94efd380908c9b824/testing/report.html)

PR #430, opened 2022-10-18
CR publication prep and IR finalization

mmccool:
- Updates to CR:
    - small wording change in at-risk section: "deployments" -> "deployment policy"
    - to be consistent with versions of text used in Architecture and Thing Description
- Updates to test results
   - CoAP and CoRE-RD improvements, several at-risk items removed

To do:
- Update IR, dates, etc. consistent with Architecture and TD IRs
- Add an Ed Note regarding status of IANA request (to sotd, to IANA considerations, AND to section 6.3) saying exact name is subject to pending IANA registration and the name is subject to change, but not the rest of the normative content.


k-toumura:
- [Rendered version of Overview.html (1a1260c)](https://cdn.statically.io/gh/w3c/wot-discovery/1a1260cf904ee31a7de4e78ec5aa194e491579da/publication/4-cr/Overview.html)
- [Rendered version of IR (1a1260c)](https://cdn.statically.io/gh/w3c/wot-discovery/1a1260cf904ee31a7de4e78ec5aa194e491579da/publication/4-cr/testing/report.html)

mlagally:
Some review comments:

1. The description contains a broken reference to the implementation reports in Section 3.

> Implementer reports were collected through the W3C WoT Interest Group's [PlugFest](https://www.w3.org/2016/07/wot-ig-charter.html#scope) activity and collected in the GitHub repository [https://github.com/w3c/wot](https://github.com/w3c/wot/tree/master/testing/tests/2019-05/) under testing/tests/2019-05/.

2. There are no implementation descriptions contained in section 6. 

3. Section 8 would benefit from a legend that explains the meaning of  "P	F	N	T"

egekorkan:
@mlagally here are some explanations for your comments:
1. I am not entirely sure but I think that this link is a placeholder. The Discovery document will be published under this link. @mmccool correct me if I am wrong.
2. As it was also done with the TD, the implementation descriptions are in the architecture specification. This was decided quite some time ago but we can simply duplicate them into here in the CR transition. A first time reader might have the same question so we should link to the arch IR in the worst case.
3. The last bullet list item at the bullet list in the beginning of Section 8 actually explains what these letters mean. However, this text can be overlooked and it would not hurt to add more information. A simple improvement would be to add the title attribute which would show the full text when the mouse is hovered over it.

PR #436, opened 2022-11-01
Mark Size2 assertion as "pass" for RIOT WoT implementation

JKRhb:
This is a small change that hasn't been copied over from the wot-testing repo (see https://github.com/w3c/wot-testing/pull/472).


mmccool:
Huh.  Could have sworn I fixed that.  But ok, will merge and update IR.

mmccool:
So, this does not change the at-risk status of that assertion, but now we only need one more implementation for it.

PR #438, opened 2022-11-01
List ALL at-risk assertions in sotd

mmccool:
To avoid delays in CR transition, list ALL at-risk assertions in sotd section (this PR adds a list of the S&P ones).


mmccool:
Main call 2022-11-02: agreed to merge once conflicts resolved.

PR #442, opened 2022-11-25
Update Implementation Report (indices, assertion text in template.csv)

mmccool:
- Add indices and assertion text to template.csv
- Order assertions by appearance in the text
- Read assertions from static version in publication/4-cr/Overview.html to resolve ReSpec rendering e.g. of section refs and citations

See also [TD PR #1752](https://github.com/w3c/wot-thing-description/pull/1752).  The tooling has also been updated to deal with other situations that are not used in this spec.

Note that the diffs for the CSV and CSS files seem large.  However, while they are reordered and have new columns, they represent the same set of assertions.


mmccool:
Merging, but noted it would be useful if index was in a separate column.  I will consider for the next update to the tooling.

PR #447, opened 2022-12-14
Update Implementation Report - Dec 2022 Testfest

mmccool:
New results for node-wot, VIAMEE, wothive


mmccool:
Going to defer merging this until I fix the results for wot-hive.  It looks like some previously reported automatic results got deleted, which would explain the reversion of the SPARQL results.  After I fix this I will update this PR and merge.

mmccool:
Fixed wothive results, resolving reversions, merging as agreed.

PR #451, opened 2022-12-21
DID context fix

### LINKED ISSUE #449: DID Implementation Fix
DID is not marked as at-risk but unfortunately the DID spec requires an `@context` which we do not have - so while our implementations pass, they are incorrect according to the DID spec.  Do we hold up the CR?  Revert these to at-risk?

mmccool:
Resolves #449 

- Updates examples to include `@context`
- Adds normative reference to did-spec-registry
- Adds assertion about mandatory use of appropriate context URL
- Updates IR, CR candidate files, etc. consistently, marking DID section as at risk, making sure at-risk highlighting for the new assertion is correct, listing DID in sotd as at-risk and why
- Adds static.html at top level to use as a reference for ir generation

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/wot-discovery/pull/451.html" title="Last updated on Dec 21, 2022, 5:45 PM UTC (44adf35)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-discovery/451/d8b682f...44adf35.html" title="Last updated on Dec 21, 2022, 5:45 PM UTC (44adf35)">Diff</a>


k-toumura (approved):
Thanks for the update.

We have little implementation experience, so I would agree to make this section informative if further investigation is needed.

mmccool:
This PR seems to touch a lot of other related files e.g. for testing inputs, by accident.  Will try to fix, but for now focus just on the change to index.html.  Note that even just making this section informative will require another CR, since it was NOT marked as at-risk when the CR was published.  We should see who did the implementations and how much work it would be to update them.  The change is not that big, but the DID document needs to be updated to include the `@context` reference.

mmccool:
Will close without merging.  Will be making new PR that only updates index.html in a draft-CR2 branch.

PR #459, opened 2023-01-23
Update Implementation Report - Jan 2023 

mmccool:
This captures a change to how the DNS-SD values were tabulated, which renumbered some of the assertions.  Also some white space changes (due to a script issue) which however do not change any content (and which I will fix in a later PR).

***
<a href="https://pr-preview.s3.amazonaws.com/mmccool/wot-discovery/pull/459.html" title="Last updated on Jan 23, 2023, 2:35 PM UTC (4bd8914)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-discovery/459/02b49fa...mmccool:4bd8914.html" title="Last updated on Jan 23, 2023, 2:35 PM UTC (4bd8914)">Diff</a>


mmccool:
Not going to merge, need to clean up other things first.  Also, probably should be careful to use a branch for the draft CR2 in case we decide NOT to proceed with that, since we would need to revert to a version without those changes for the IR.

PR #462, opened 2023-01-23
CR Update - DID and DNS-SD fixes

### LINKED ISSUE #403: Refine DNS-SD Name Requirements and IANA Consideration
- What exactly needs to be registered?  
     - Just _wot under SRV and TXT?  
     - Do we have to register PTR?  
     - Do we have to register _directory?
- Remove editor's notes (or comment them out), at this point they are confusing
- Then update IANA registration request; for now we will tell them we are working on it (@ashimura will respond to IANA; note: actual registration can be post-CR, but fixing the text in our doc should be pre-CR)

Should also ask PLH - has anyone else done DNS-SD registrations?  (@ashimura will ask)

mmccool:
Resolve #403 

<strike>WIP: This PR should NOT be merged until we have a group resolution to proceed with a second CR for WoT Discovery.  At this point it is a draft candidate.  I have made this PR because (a) once/if we do have a resolution to proceed then the draft-CR2 branch should be merged into the working draft (b) it results in a link to render the draft and show a diff.</strike>

Fixes to DID and DNS-SD Introduction methods.  As discussed below, I have confirmed with @plehegar that these can be considered editorial and/or bug fixes, and so can be merged into the current editor's draft and included in the Proposed Recommendation *without* a new CR.

To do: 
* update the change log.
* review from DID

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/wot-discovery/pull/462.html" title="Last updated on Feb 20, 2023, 3:50 PM UTC (a1f409d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-discovery/462/02b49fa...a1f409d.html" title="Last updated on Feb 20, 2023, 3:50 PM UTC (a1f409d)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -681,35 +681,54 @@ img.wot-diagram {
                 <code>WotThing</code> or <code>WotDirectory</code>, respectively, in the DID Document that 
                 the TD's identifier resolves to.</span>
             </p>
+            <p><span class="rfc2119-assertion" id="introduction-did-service-context">In order to define Service Endpoints
+                for WoT Discovery, the DID Document obtained by resolving the DID of a <a>Thing</a> or <a>Thing Description Directory</a>
+                MUST include the URL <code>https://www.w3.org/2021/wot/discovery/did</code> in its <code>@context</code> 
+                [[did-spec-registries]].
+            </span>
+            </p>
             <p><span class="rfc2119-assertion" id="introduction-did-service-endpoint">
                 If the DID Document obtained by resolving the DID of a <a>Thing</a> or <a>Thing Description Directory</a> 
                 contains a Service Endpoint of type <code>WotThing</code> or <code>WotDirectory</code>, respectively,
                 then this Service Endpoint MUST refer to the <a>TD</a> describing that Thing 
                 (when using the <code>WotThing</code> service name) 
 		            or the <a>TD</a> describing that <a>Thing Description Directory</a> (when using the <code>WotDirectory</code> service name),
-                respectively.</span>
+                respectively
+                [[did-spec-registries]].</span>
             </p>
             <aside class="example" title="A Example Service Endpoint in a DID Document - WotThing">
                 <pre>
                     {
+                        "@context":[
+                            "https://www.w3.org/ns/did/v1",
+                            "https://www.w3.org/2021/wot/discovery/did"
+                        ],
+                        ...
                         "service": [{
                             "id": "did:example:wotdiscoveryexample#td",
                             "type": "WotThing",
                             "serviceEndpoint":
                                 "https://wot.example.com/.well-known/wot"
                         }]
+                        ...
                     }
                 </pre>
             </aside>
             <aside class="example" title="A Example Service Endpoint in a DID Document - WotDirectory">
                 <pre>
                     {
+                        "@context":[
+                            "https://www.w3.org/ns/did/v1",
+                            "https://www.w3.org/2021/wot/discovery/did"
+                        ],
+                        ...
                         "service": [{
                             "id": "did:example:wotdiscoveryexample#tdd",
                             "type": "WotDirectory",
                             "serviceEndpoint":
                                 "https://wot.example.com/tdd"
                         }]
+                        ...
                     }
                 </pre>
             </aside>
--- a/index.html
+++ b/index.html
@@ -579,8 +579,6 @@ img.wot-diagram {
         </section>
         <section id="introduction-dns-sd-sec" class="normative">
             <h2>DNS-Based Service Discovery</h2>
-            <p class="ednote">The DNS-SD service names are still subject to IANA approval.  If necessary they may have to be
-            changed prior to PR if IANA requires different names.</p>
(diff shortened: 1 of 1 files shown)


mmccool:
I met with @plehegar and he agreed that in principle we can merge these fixes into our current CR round, and include them in the PR without an additional CR transition.  I have updated the title of this PR accordingly.  It's still WIP as I need to update the change log.

I also plan to ask DID for a sanity check and we will have to redo the testing of these features since the content of the DID document will change.

mmccool:
Still would like to check with DID and update the change log.  I don't expect any other problems but waiting another week will not impact the schedule.

What WILL impact the schedule is testing the updated DID implementation, e.g. to include an `@context` in the DID documents used.  Both implementations were by Hitachi: hitachi-esf-idf and hitachi-node-red.  @k-toumura has said he's updated the DID documents already, and wot-testing has been updated, as the implementations have been updated to reflect this change.  However, one remaining issue is that the ontology file has not yet been defined and the provided URL does not resolve (gives a 404).

mmccool:
Regarding the ontology, there are two issues: generating and testing the ontology file, and allocating the URLs (and making them consistent with the namespaces allocated for TD).  The TD spec uses https://www.w3.org/2022/wot/td/v1.1 (see https://w3c.github.io/wot-thing-description/#namespaces), the Discovery spec currently uses https://www.w3.org/2021/wot/discovery.

Proposal:
- Align with (proposed) https://www.w3.org/2022/wot namespace
- Allocate a sub-space in there for discovery: https://www.w3.org/2022/wot/discovery
- Update the CR to use the following:
     - https://www.w3.org/2022/wot/discovery/core - API ontology
     - https://www.w3.org/2022/wot/discovery/did - DID document ontology

index.html
```diff
@@ -681,35 +697,54 @@ <h2>DID Documents</h2>
                 <code>WotThing</code> or <code>WotDirectory</code>, respectively, in the DID Document that 
                 the TD's identifier resolves to.</span>
             </p>
+            <p><span class="rfc2119-assertion" id="introduction-did-service-context">In order to define Service Endpoints
+                for WoT Discovery, the DID Document obtained by resolving the DID of a <a>Thing</a> or <a>Thing Description Directory</a>
+                MUST include the URL <code>https://www.w3.org/2021/wot/discovery/did</code> in its <code>@context</code>
```
msporny, line 702: This URL doesn't resolve, but I expect that will be fixed in time.

You might also want to specify that the URL must exist after the 1st array entry in the `@context` field. It should be obvious for those in the know, but it might not be... perhaps an example would help (I don't know if you already have one in the spec since I'm reviewing the PR instead of the human readable version).

msporny (approved):

index.html
mmccool: Thanks for your feedback.  I'm not really happy with the year-based-URLs either since we then ALSO have version numbers, and the year we allocated no longer aligns with the date of publication of our spec, etc. etc.  However, our main goal at present is to at least be consistent with our current set of publications.  Our other constraint is that we were previously using year-based URLs, some of which we will still be using, so...

mmccool: Anyway, we are working on fixing the URL and having it resolve to an appropriate ontology definition.

index.html
```diff
@@ -631,8 +630,25 @@ <h2>DNS-Based Service Discovery</h2>
...
+                    </p>
+                    <dl>
+                        <dt><code>td</code></dt>
+                        <dd>Absolute pathname of the Thing Description of the Thing or Thing Description of the <a>Thing Description Directory</a>.</dd>
+                        <dt><code>type</code></dt>
+                        <dd>Type of the Thing Description, i.e. <code>Thing</code> or <code>Directory</code>.
+                        If omitted, the type is assumed to be <code>Thing</code>.</dd>
+                        <dt><code>scheme</code></dt>
+                        <dd>Scheme part of URL.  One of the following values can be specified, with the standard registered URI interpretations [[RFC7595]]: <code>coap</code> (CoAP over UDP) or <code>coaps</code> (CoAP over DTLS/UDP).
+                        If omitted, the scheme is assumed to be <code>coap</code>.</dd>
```
JKRhb, line 651: One thing I noticed when implementing DNS-SD discovery was that there is currently no way to include a port in the TXT record, so you have to rely on the default ports for the indicated/implied URI scheme. Would it make sense to add another optional attribute for the port here?

farshidtz: Maybe I misunderstood the issue, but there is a dedicated entry for port that goes into the service registration. It doesn't need to be passed as a TXT record; see [here](https://www.rfc-editor.org/rfc/rfc6763#section-6) and other parts of the spec.

For example, see  https://github.com/TinyIoT/thing-directory/wiki/Discovery-with-DNS-SD (may not be compliant with the current Discovery spec but you can see the custom port).

The user manuals for registering a service via dns-sd and avahi CLI apps could also be used as reference:
- https://man.netbsd.org/dns-sd.1
- https://linux.die.net/man/1/avahi-publish-service

PR #469, opened 2023-03-27
Add workaround for mandatory response contentType

### LINKED ISSUE #465: TDD Thing Model is missing `contentType` in `response` objects
Experimenting with discovery from TDDs, I noticed that a TDD implementation that adheres to the currently defined Thing Model for the API specification ends up generating a Thing Description that is not valid according to the TD specification. That is because a form like the following does not define a `contentType` for its `response` member, causing an instantiated TD to be invalid if no `contentType` is explicitly added by the TDD:

```js
{
    "href": "/things/{id}",
    "htv:methodName": "PUT",
    "response": {
        "htv:statusCodeValue": 201
        // No contentType defined!
    },
    "additionalResponses": [
        {
            ...
        }
    ]
}
```

As also discussed in https://github.com/w3c/wot-thing-description/pull/1776, having no `contentType` seems reasonable to me here – however, if a discovery validates a TDD's Thing Description, then it is rejected in its current form.

Would it be a possible workaround to add `application/octet-stream` for cases like these? Or is there maybe a better alternative?

JKRhb:
Resolves #465.

***
<a href="https://pr-preview.s3.amazonaws.com/JKRhb/wot-discovery/pull/469.html" title="Last updated on Apr 3, 2023, 2:31 PM UTC (8547b40)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-discovery/469/02b49fa...JKRhb:8547b40.html" title="Last updated on Apr 3, 2023, 2:31 PM UTC (8547b40)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -2453,6 +2453,15 @@ img.wot-diagram {
                     given in this Thing Model
                     are not normative and are provided for information only.
                     </p>
+                    <p>
+                        Note that a <code>contentType</code> is required by the WoT TD 1.1
+                        specification [[wot-thing-description11]] for instances of the
+                        <code>ExpectedResponse</code> class (i.e., the JSON object
+                        corresponding to the key <code>response</code> in a form),
+                        even when the body of the HTTP response is supposed to be empty.
+                        In this case, <code>application/octet-stream</code> can be used to
+                        ensure that the TD generated from the TM below is valid.
+                    </p>
                     <pre class="advisement json"
                          data-include='directory.tm.json'></pre>
 
--- a/index.html
+++ b/index.html
@@ -2459,7 +2459,7 @@ img.wot-diagram {
                         <code>ExpectedResponse</code> class (i.e., the JSON object
                         corresponding to the key <code>response</code> in a form),
                         even when the body of the HTTP response is supposed to be empty.
-                        In this case, <code>application/octet-stream</code> can be used to
+                        In this case, <code>application/x-empty</code> can be used to
                         ensure that the TD generated from the TM below is valid.
                     </p>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -2453,6 +2453,15 @@ <h4>API Specification (Thing Model)</h1>
                     given in this Thing Model
                     are not normative and are provided for information only.
                     </p>
+                    <p>
+                        Note that a <code>contentType</code> is required by the WoT TD 1.1
+                        specification [[wot-thing-description11]] for instances of the
+                        <code>ExpectedResponse</code> class (i.e., the JSON object
+                        corresponding to the key <code>response</code> in a form),
+                        even when the body of the HTTP response is supposed to be empty.
+                        In this case, <code>application/octet-stream</code> can be used to
```
farshidtz, line 2462: I don't think `application/octet-stream` is an appropriate content type to use here.  That is used for binary data to be placed in a file. See:

> The "octet-stream" subtype is used to indicate that a body contains
arbitrary binary data.
> ...
> The recommended action for an implementation that receives an
"application/octet-stream" entity is to simply offer to put the data
in a file, with any Content-Transfer-Encoding undone, or perhaps to
use it as input to a user-specified process.
https://www.iana.org/assignments/media-types/application/octet-stream

The following is not ideal, but at least it isn't indicating something false:
```suggestion
                        In this case, <code>application/x-empty</code> can be used to
```

JKRhb: I'd be okay with this change – I suppose there are no optional solutions for this problem :/ What do you think, @mmccool?

PR #470, opened 2023-03-27
Rework RFC 9175 assertion and make it informative

### LINKED ISSUE #466: Removing the RFC 9175 assertion?
One assertion in the "Amplification and Distributed Denial of Service" section currently reads as follows:

> If support for multicast is required, in the case of CoAP, the recommendations made in [[RFC9175](https://www.w3.org/TR/wot-discovery/#bib-rfc9175)] SHOULD be applied.

Revisiting this assertion, I noticed that this wording does not make that much sense, since RFC 9175 only contains an explicit recommendation for multicast in the case of OSCORE-protected group communication. Furthermore, it already contains this assertion, making the assertion in the Discovery specification a bit redundant:

> A CoAP server SHOULD mitigate potential amplification attacks by responding to unauthenticated clients with 4.01 (Unauthorized) including an Echo option, as described in item 3 in [Section 2.4](https://datatracker.ietf.org/doc/html/rfc9175#echo-app) of RFC 9175.

Therefore, I was wondering if the reference to RFC 9175 should be dropped completely or reworded to an informative statement, possibly referring to [`draft-irtf-t2trg-amplification-attacks`](https://datatracker.ietf.org/doc/html/draft-irtf-t2trg-amplification-attacks) (where possible amplification attacks using CoAP and multicast are illustrated) instead.

I suppose this issue could also be discussed in the next meeting of the security TF.

JKRhb:
This PR resolves #466, by changing the cited reference from RFC 9175 to `draft-irtf-t2trg-amplification-attacks` (which is the document that should have been cited in the first place) and "downgrading" the statement from normative to informative. The PR also fixes the actual content of the statement, since the cited document does not make any real recommendations regarding the use of CoAP with multicast; instead, it mostly states that 

> [f]or multicast requests, anti-amplification limits and the Echo Option do not really work unless the number of servers sending responses is known. Even if the responses have the same size as the request, the amplification factor from m servers is m, where m is typically unknown.

However, I am not sure if Group OSCORE could be a potential solution here or rather if the authors are implying that it could be one (CC @j1y3p4rk). In any case, though, this PR should already fix the main issue that is currently present in this part of the document.

***
<a href="https://pr-preview.s3.amazonaws.com/JKRhb/wot-discovery/pull/470.html" title="Last updated on Mar 28, 2023, 5:42 PM UTC (b4678b9)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-discovery/470/209c083...JKRhb:b4678b9.html" title="Last updated on Mar 28, 2023, 5:42 PM UTC (b4678b9)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -2620,14 +2620,10 @@ img.wot-diagram {
 	        <span class="rfc2119-assertion" id="sec-tdd-intro-no-multicast">
 		Open implementations of Introduction mechanisms SHOULD NOT respond to multicast
 		requests unless this is absolutely required by the protocol.</span>
-	        <span class="rfc2119-assertion" id="sec-tdd-intro-if-multicast-required">
 		If support for multicast is required,
-		in the case of CoAP, the recommendations made in [[RFC9175]] SHOULD be
-		applied.</span>
-		Note however that in the case of discovery the number of servers that
-		might respond to a multicast request will generally not be known in
-		advance, in which case the mitigations proposed in [[RFC9175]] may
-		not be effective.
+		in the case of CoAP, the observations made in [[AMPLIFICATION-ATTACKS]] are relevant,
+		since the number of servers that might respond to a multicast request during
+		discovery will generally not be known in advance.
 		</li>
 		<li>
 		Limit the size of responses to the minimum.  


mmccool (approved):
Looks good.  I see you also updated the following note about the issue with the number of servers not being known in advance.

JKRhb:
(Sorry for the noise.) Reformatted the bibliography entry and fixed the cite key in the actual text.

PR #471, opened 2023-03-27
Update context URLs

### LINKED ISSUE #463: Align Context URLs with TD specification
Currently the ontology URLs in the doc are not aligned with those use/proposed for TD 1.1.  Propose we do the following:
- Align with (proposed) https://www.w3.org/2022/wot namespace from TD 1.1 spec
- Allocate a sub-space in there for discovery: https://www.w3.org/2022/wot/discovery
- Update the CR to use the following:
     - https://www.w3.org/2022/wot/discovery/core - API ontology
     - https://www.w3.org/2022/wot/discovery/did - DID document ontology

Note the current discovery spec is very similar but uses 2021 and does not use "core".

mmccool:
Resolves points 1b, 2, 3 of [#463.](https://github.com/w3c/wot-discovery/issues/463#issuecomment-1485272509), copied here for reference:
1. To be consistent with the TD spec, use "https://www.w3.org/2022/wot/discovery/did" for the DID @context reference.
    a. Update the Discovery doc
    b. Update the DID PR at https://github.com/w3c/did-spec-registries/pull/486)
2. Also update the URL for the TDD vocabulary to "https://www.w3.org/2022/wot/discovery/tdd" (this changes both the year, to be consistent with the TD spec, and adds "tdd" so that both ontology files are at the same level).
3. Update all examples, etc. to use the correct URLs.

The PR against the DID registry has also been updated to use 2022.

Other changes:
- Got rid of github.io urls generally, in preparation for PR transition, replacing all non-DID contexts with the https://www.w3.org/2022/wot/discovery/tdd.  Note that this context URL is also used for collections and event reports.  We *could* use different context files for these but I did not want to deviate too far from the original intent, given that CR/PR is supposed to just be editorial changes.
- Updated the TD @context references to the one defined in the TD1.1 specification.
- Made sure to use https URLs generally and consistently for context files.

Note that some additional steps are needed to fully resolve #463 - in particular we need to update implementations to use the new URLs, need to revalidate tests, and need to make the relevant ontologies available at the given URLs.

***
<a href="https://pr-preview.s3.amazonaws.com/mmccool/wot-discovery/pull/471.html" title="Last updated on May 22, 2023, 3:25 PM UTC (77e8f96)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-discovery/471/209c083...mmccool:77e8f96.html" title="Last updated on May 22, 2023, 3:25 PM UTC (77e8f96)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -2082,7 +2082,7 @@ img.wot-diagram {
 					title="Alternative payload format including pagination information">
 					<pre>
 						{
-							"@context": "https://www.w3.org/2022/wot/discovery/tdd",
+							"@context": "https://www.w3.org/2022/wot/discovery",
 							"@type": "ThingCollection",
 							"total": 12,
 							"members": [
--- a/index.html
+++ b/index.html
@@ -699,7 +699,7 @@ img.wot-diagram {
             </p>
             <p><span class="rfc2119-assertion" id="introduction-did-service-context">In order to define Service Endpoints
                 for WoT Discovery, the DID Document obtained by resolving the DID of a <a>Thing</a> or <a>Thing Description Directory</a>
-                MUST include the URL <code>https://www.w3.org/2021/wot/discovery/did</code> in its <code>@context</code> 
+                MUST include the URL <code>https://www.w3.org/2022/wot/discovery/did</code> in its <code>@context</code> 
                 [[did-spec-registries]].
             </span>
             </p>
@@ -717,7 +717,7 @@ img.wot-diagram {
                     {
                         "@context":[
                             "https://www.w3.org/ns/did/v1",
-                            "https://www.w3.org/2021/wot/discovery/did"
+                            "https://www.w3.org/2022/wot/discovery/did"
                         ],
                         ...
                         "service": [{
@@ -735,7 +735,7 @@ img.wot-diagram {
                     {
                         "@context":[
                             "https://www.w3.org/ns/did/v1",
-                            "https://www.w3.org/2021/wot/discovery/did"
+                            "https://www.w3.org/2022/wot/discovery/did"
                         ],
                         ...
                         "service": [{
@@ -833,7 +833,7 @@ img.wot-diagram {
 
                     <span class="rfc2119-assertion" id="exploration-directory-description-type"> 
                         A TD which describes a Thing Description Directory instance MUST use type `ThingDirectory` from the
-                        discovery context or URI `https://www.w3.org/2021/wot/discovery#ThingDirectory`.</span>
+                        discovery context or URI `https://www.w3.org/2022/wot/discovery/tdd#ThingDirectory`.</span>
                     <p>
                         A TD of this class can be derived from Directory's Thing Model; see [[[#directory-api-spec]]].
                     </p>
@@ -847,7 +847,7 @@ img.wot-diagram {
 
                     <span class="rfc2119-assertion" id="exploration-link-description-type"> 
                         A TD which describes a reference to another TD MUST use type `ThingLink` from the
-                        discovery context or URI `https://www.w3.org/2021/wot/discovery#ThingLink`.</span>
+                        discovery context or URI `https://www.w3.org/2022/wot/discovery/tdd#ThingLink`.</span>
                     <span class="rfc2119-assertion" id="exploration-link-description-link"> 
                         A Thing Link MUST define the referenced TD as a Link with
                         `describedby` link relation type, `application/td+json` media type
@@ -862,8 +862,8 @@ img.wot-diagram {
                         <pre>
                             {
                                 "@context": [
-                                    "http://www.w3.org/ns/td",
-                                    "https://w3c.github.io/wot-discovery/context/discovery-context.jsonld"
+                                    "https://www.w3.org/2022/wot/td/v1.1",
+                                    "https://www.w3.org/2022/wot/discovery/tdd"
                                 ],
                                 "@type": "ThingLink",
                                 "id": "urn:example:link",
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -1227,7 +1227,7 @@ <h4>Registration Information</h4>
                     use within <a>TDs</a> that embed or reference the Discovery context. 
                     Note that only an <a>Enriched TD</a> embeds the registration information. 
 		            <span class="rfc2119-assertion" id="tdd-context-injection">An <a>Enriched TD</a>
-                        MUST contain in its @context the URI <code>https://w3c.github.io/wot-discovery/context/discovery-context.jsonld</code>.</span>
+                        MUST contain in its @context the URI <code>https://www.w3.org/2022/wot/discovery/tdd</code>.</span>
```
mmccool, line 1230:
```suggestion
                        MUST contain in its @context the URI <code>https://www.w3.org/2022/wot/discovery/enriched</code>.</span>
```

index.html
```diff
@@ -1683,8 +1683,8 @@ <h4>Retrieval</h4>
                                 <pre>
                                     {
                                         "@context": [
-                                            "http://www.w3.org/ns/td",
-                                            "https://w3c.github.io/wot-discovery/context/discovery-context.jsonld"
+                                            "https://www.w3.org/2022/wot/td/v1.1",
+                                            "https://www.w3.org/2022/wot/discovery/tdd"
```
mmccool, line 1687:
```suggestion
                                            "https://www.w3.org/2022/wot/discovery/enriched"
```

index.html
```diff
@@ -1717,8 +1717,8 @@ <h4>Retrieval</h4>
...
```
mmccool, line 1721:
```suggestion
                                            "https://www.w3.org/2022/wot/discovery/enriched"
```

index.html
```diff
@@ -1749,8 +1749,8 @@ <h4>Retrieval</h4>
...
```
mmccool, line 1753:
```suggestion
                                            "https://www.w3.org/2022/wot/discovery/enriched"
```

index.html
```diff
@@ -2289,8 +2289,8 @@ <h4>Events API</h4>
                                             event: thing_created
                                             data: {
                                             data:     "@context": [
-                                            data:         "https://www.w3.org/2019/wot/td/v1",
-                                            data:         "https://w3c.github.io/wot-discovery/context/discovery-context.jsonld"
+                                            data:         "https://www.w3.org/2022/wot/td/v1.1",
+                                            data:         "https://www.w3.org/2022/wot/discovery/tdd"
```
mmccool, line 2293:
```suggestion
                                            data:         "https://www.w3.org/2022/wot/discovery/enriched"
```

mmccool:
Ugh, it seems we have some other issues, i.e. ontologies are used for three purposes: collections (returned for pagination), enriched TDs, and TDs of TDDs.  Ideally we would have three separate ontologies for these, but our assertions are not structured right for that, in particular we don't have an assertion for the "collection" usage, only an example.

Resolution: put all usages except DID into one ontology.  However, the name cannot be "tdd" since that is not the only use case.   We discussed it and decided to use "discovery-did" for the ontology to be used in DID and just "discovery" for everything else.  Using a hyphen avoids any possible issue with a URL being used for both a directory and a file.

I'll update the PR...

index.html
```diff
@@ -699,7 +699,7 @@ <h2>DID Documents</h2>
             </p>
             <p><span class="rfc2119-assertion" id="introduction-did-service-context">In order to define Service Endpoints
                 for WoT Discovery, the DID Document obtained by resolving the DID of a <a>Thing</a> or <a>Thing Description Directory</a>
-                MUST include the URL <code>https://www.w3.org/2021/wot/discovery/did</code> in its <code>@context</code> 
+                MUST include the URL <code>https://www.w3.org/2022/wot/discovery/did</code> in its <code>@context</code>
```
mmccool, line 702:
```suggestion
                MUST include the URL <code>https://www.w3.org/2022/wot/discovery-did</code> in its <code>@context</code> 
```

index.html
```diff
@@ -717,7 +717,7 @@ <h2>DID Documents</h2>
                     {
                         "@context":[
                             "https://www.w3.org/ns/did/v1",
-                            "https://www.w3.org/2021/wot/discovery/did"
+                            "https://www.w3.org/2022/wot/discovery/did"
```
mmccool, line 720:
```suggestion
                            "https://www.w3.org/2022/wot/discovery-did"
```

index.html
```diff
@@ -735,7 +735,7 @@ <h2>DID Documents</h2>
...
```
mmccool, line 738:
```suggestion
                            "https://www.w3.org/2022/wot/discovery-did"
```

index.html
```diff
@@ -833,7 +833,7 @@ <h4>`ThingDirectory`</h4>
 
                     <span class="rfc2119-assertion" id="exploration-directory-description-type"> 
                         A TD which describes a Thing Description Directory instance MUST use type `ThingDirectory` from the
-                        discovery context or URI `https://www.w3.org/2021/wot/discovery#ThingDirectory`.</span>
+                        discovery context or URI `https://www.w3.org/2022/wot/discovery/tdd#ThingDirectory`.</span>
```
mmccool, line 836:
```suggestion
                        discovery context or URI `https://www.w3.org/2022/wot/discovery#ThingDirectory`.</span>
```

index.html
```diff
@@ -862,8 +862,8 @@ <h4>`ThingLink`</h4>
                         <pre>
                             {
                                 "@context": [
-                                    "http://www.w3.org/ns/td",
-                                    "https://w3c.github.io/wot-discovery/context/discovery-context.jsonld"
+                                    "https://www.w3.org/2022/wot/td/v1.1",
+                                    "https://www.w3.org/2022/wot/discovery/tdd"
```
mmccool, line 866:
```suggestion
                                    "https://www.w3.org/2022/wot/discovery"
```

index.html
```diff
@@ -2082,7 +2082,7 @@ <h4>Listing</h4>
 					title="Alternative payload format including pagination information">
 					<pre>
 						{
-							"@context": "https://w3c.github.io/wot-discovery/context/discovery-context.jsonld",
+							"@context": "https://www.w3.org/2022/wot/discovery/tdd",
```
mmccool, line 2085:
```suggestion
							"@context": "https://www.w3.org/2022/wot/discovery",
```

mmccool:
OK, updated URLs, also updated PR against DID registry.  Please review... will discuss in meeting on April 10.

PR #473, opened 2023-04-10
April 2023 Implementation Result updates

mmccool:
- Fix wothive result.  This was previously commited in "wothive/manual.csv" but not added to "wothive.csv" in the Results directory.  This removes "tdd-things-list-pagination-collection" from the set of at-risk assertions.
- Updated node-wot results.  This removes "exploration-server-http-alternate-content" and "sec-tdd-intro-no-observe" from the set of at-risk assertions, and provides one implementation for "sec-tdd-intro-no-multicast".
- Add wot-coap-dns-sd.csv results file - for some reason it was not checked in.

Will continue doing updates until the Discovery dev meeting on April 17: we want to trim down the at-risk list as much as possible by then.  I will also be updating the Dev meeting slides to match (slides will be hidden when issue resolved but not removed).


mmccool:
Merging so template.csv files are ready for Testfest.  Will start another PR for new changes.

Issue #480, opened 2023-05-16
sort_by should be a JSON pointer


mmccool:
- The value of sort_by should be a JSON pointer so it can refer to any field, including nested ones
- However JSON pointers need to start with a /, generally
- Example does not do this...

farshidtz:
I agree that the current spec isn't clear on what to do with nested fields and support choosing JSON Pointer as **the only notation** in Discovery v2. However, we need to make sure such a change doesn't add implementation burden for servers which use relational or semantic DBs.

PR #481, opened 2023-05-16
Updated IR - May 2023 - wot-ha results on sorting and pagination

mmccool:
- new results after implementation of sorting and pagination in wot-ha
- resolves at-risk status of tdd-things-list-pagination-order, tdd-things-list-pagination-orderable, tdd-things-list-pagination-order-nextlink, and tdd-things-list-pagination-order-utf-8

Note 1: I noticed we did not specify in the spec what valid values of "sort_by" should be.  Since we are referring to a JSON object, they probably should be JSON pointers, which should generally start with a slash, like "/id" or "/title".   BUT our example does not have a slash... created an issue: https://github.com/w3c/wot-discovery/issues/480.  My implementation uses JSON pointers for this parameter.
Note 2: apparently you can only duplicate fields in HTTP Headers if they can be converted to comma-separated lists.  The spec requires Link to be duplicated (you need both canonical and next links), but my tools (node.js) won't let me have multiple headers with the same field name, so I used a comma-separated list.


mmccool:
Some related concerns, maybe deal with in Discovery 2.0:
- What happens if you want to sort by titles but some TDs don't have them?  Do you return an error, or just use some default value that clumps all the undefined values together?  Do you sort the undefined one on a secondary key?  My implementation gives an error in this case but this is arguably the least useful way to handle this.
- What error value should be returned if a query is malformed?  I use 400 in my implementation (Bad Request).
- What error value should be returned if a sort is not possible, e.g. if a sort_by refers to an object?  I use 400 in my implementation (Bad Request).  Note the related assertion only says "reject requests", it does not say how.
- Dates are stored as strings but we may want to have some special rules for sorting them.  Fortunately ISO dates are designed to be lexicographically sortable, but they all have to be in the same timezone, e.g. UTC (which we have other assertions about...).

PR #485, opened 2023-05-19
Resolve exploration-secboot-oauth2-flows

mmccool:
- Resolve at-risk exploration-secboot-oauth2-flows due to insufficient implementations
- Need to remove, and since not supporting redirections with 3xx error codes breaks the only OAuth2 flow (code) that works in this case, remove other mentions of OAuth2
- Commented out as we really *should* have it and it should be a priority to include in the next version of Discovery/onboarding
- Note that could still use the "client" flow - in this case you know the auth server already, and can go get a Bearer token and then connect that way.

***
<a href="https://pr-preview.s3.amazonaws.com/mmccool/wot-discovery/pull/485.html" title="Last updated on May 19, 2023, 4:41 PM UTC (231de67)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-discovery/485/25bf37a...mmccool:231de67.html" title="Last updated on May 19, 2023, 4:41 PM UTC (231de67)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -988,12 +988,14 @@ img.wot-diagram {
                 are several authentication schemes registered with IANA.</a>
                 However, not all of these are in wide use, some are experimental, and there is only
                 partial overlap with the schemes supported by TDs.
+<!-- See below - need to take out due to insufficient implementation experience.  Remove reference to OAuth1 as well.
                 Also, note that the <code>oauth</code> scheme in the IANA registration refers
                 to OAuth1, which is deprecated, so it should not be used.
                 The relevant OAuth2 flow, the <code>code</code> flow, instead of a 401 response
                 begins with a redirection to an authentication
                 server, eventually resulting in credentials (bearer tokens in the case of WoT)
                 that can be used for access.
+-->
                 </p>
                 <p>
                 Given these considerations,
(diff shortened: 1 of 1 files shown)


mmccool:
@benfrancis Hmm, you are correct.  In fact there was one big assertion covering both 401 and 300x errors.  So the entire thing needs to be made informative.   I will close this PR as it does the wrong thing; will have to redo.

benfrancis:
@mmccool Surely, technically, as long as a server responded with _either_ 401 response, _or_ a 302/304 response, the original assertion #exploration-secboot-401 was satisfied and therefore we do have implementations. Its only #exploration-secboot-oauth2-flows (specifically about OAuth2) which should be at risk and needs to be downgraded. The wording I proposed was just another approach to removing mention of OAuth2 from the first assertion (as you seemed to want to do), without making OAuth2 impossible. An alternative would be to just leave #exploration-secboot-401 as it is, but I personally think it would be neater to generalise it and downgrade the other two assertions.

Please also note that the WoT Profiles specification references this section:

> Conformant Things which require authentication in order to retrieve their Thing Description MUST implement security bootstrapping, as defined in [Security Bootstrapping](https://w3c.github.io/wot-discovery/#exploration-secboot) in the WoT Discovery [[wot-discovery](https://w3c.github.io/wot-profile/#bib-wot-discovery)] specification.

If the normative assertions are removed it would make that constraint a bit meaningless.

mmccool:
One constraint we have is that we can't remove assertions that were NOT at risk going into CR.  Fortunately, both [exploration-secboot-auth](https://w3c.github.io/wot-discovery/#exploration-secboot-auth) and [exploration-secboot-oauth2-flows](https://w3c.github.io/wot-discovery/#exploration-secboot-oauth2-flows) were at risk going into CR.  So your first thought of making them both informative would work and would be consistent.  We did get some testing results for exploration-secboot-auth but we can still make it informative with a group resolution.   There are other assertions that are relevant, but if we interpret the "or" in [exploration-secboot-401](https://w3c.github.io/wot-discovery/#exploration-secboot-401) to mean "either of these conditions would satisfy the assertion" we can leave it as normative.  The "exploration-secboot-401" was also at risk going into CR but was tested; generalization might be ok as long as the generalization includes the results that were tested.

Well, let me take a stab at implementing the above...

PR #486, opened 2023-05-19
Resolve security assertions

mmccool:
- Resolve the following at-risk security assertions due to insufficient implementation:
    - sec-tdd-query-watchdog
    - sec-tdd-intro-no-multicast
    - sec-self-proxy
    - sec-tdd-throttle-queries
    - sec-tdd-limit-query-complexity
    - sec-tdd-intro-limit-response-size
    - sec-tdd-intro-throttling
- In general wording not changed
- Assertion markup commented out
- RFC2119 keywords decapitalized (SHOULD -> should, etc)

### DIFF
--- a/index.html
+++ b/index.html
@@ -2566,21 +2566,24 @@ img.wot-diagram {
                     <dt>Mitigations:</dt><dd>
                     <ul>
                     <li>
-	            <span class="rfc2119-assertion" id="sec-tdd-throttle-queries">
-                    A WoT Thing Description Directory implementation SHOULD
-                    limit the number of queries per unit time from the same requestor.</span>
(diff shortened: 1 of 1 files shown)


farshidtz (approved):

farshidtz:
Isn't it better to make the whole section informative, instead of putting together 8 informative and 4 normative assertions?

mmccool:
@farshidtz I agree that making the entire thing informative would make sense, the trouble is some of the assertions in this section were NOT at risk when we went into CR, so we technically can't take them out.  I think this (downgrading the verbs) is the simplest solution, but yeah, it's not perfect.

PR #487, opened 2023-05-19
Resolve privacy assertions

mmccool:
- Resolve the following at-risk privacy assertions due to insufficient implementation:
    - priv-loc-disable-public-directories
    - priv-loc-anonymous-tds
    - priv-loc-gen-ids
    - priv-loc-explicit-strip
    - priv-query-anon
- In general wording not changed
- Assertion markup commented out
- RFC2119 keywords decapitalized (SHOULD -> should, etc)

***
<a href="https://pr-preview.s3.amazonaws.com/mmccool/wot-discovery/pull/487.html" title="Last updated on May 19, 2023, 5:05 PM UTC (fc95537)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-discovery/487/25bf37a...mmccool:fc95537.html" title="Last updated on May 19, 2023, 5:05 PM UTC (fc95537)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -2860,10 +2860,10 @@ img.wot-diagram {
                 There are a few options to mitigate these risks:
                 <ul>
                 <li>
-	        <span class="rfc2119-assertion" id="priv-loc-disable-public-directories">
+	        <!-- <span class="rfc2119-assertion" id="priv-loc-disable-public-directories"> -->
 		To avoid location tracking and other forms of profiling, 
-    a WoT Thing associated with a person MAY 
-		disable registration with public directories.</span>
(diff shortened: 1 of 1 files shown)


farshidtz:
As with the Security Considerations, I suggest making the whole section informative instead of confusingly mixing 5 informative and 2 normative assertions.

mmccool:
@farshidtz this is a little different, since going into CR all the assertions were at risk.  However, I think taking out things that were tested should really have a group resolution.  So for consistency I think we should keep the ones that were tested as normative and do the same thing as with security.

PR #488, opened 2023-05-19
Resolve sort_by specification issue

### LINKED ISSUE #480: sort_by should be a JSON pointer

mmccool:
Resolves #480 
- Proposes change that allows field to sort_by to be specified by *either* a top-level field name *or* a JSON pointer (with RFC citation)
- an implementation could support both, since JSON pointers will always have a leading slash in this context, so if there is no slash, it can be interpreted as a field name... unless a field name starts with a slash.  Don't do that.
- Should cover all existing implementations so testing data should still be valid
- We should clean this up in the next version of the Discovery spec.  In particular we should specify that *both* should be supported by all implementations to avoid interoperability issues, or just allow one (like JSON Pointers), but that would go too far at this point (non-editorial, impacting implementations, etc).

***
<a href="https://pr-preview.s3.amazonaws.com/mmccool/wot-discovery/pull/488.html" title="Last updated on May 22, 2023, 3:07 PM UTC (4b40142)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-discovery/488/25bf37a...mmccool:4b40142.html" title="Last updated on May 22, 2023, 3:07 PM UTC (4b40142)">Diff</a>

### DIFF
--- a/directory.tm.json
+++ b/directory.tm.json
@@ -47,25 +47,11 @@
                         "collection"
                     ],
                     "default": "array"
-                },
-                "sort_by": {
-                    "title": "Comparator TD attribute for collection sorting",
-                    "type": "string",
-                    "default": "id"
-                },
-                "sort_order": {
-                    "title": "Sorting order",
-                    "type": "string",
-                    "enum": [
-                        "asc",
-                        "desc"
-                    ],
-                    "default": "asc"
                 }
             },
             "forms": [
                 {
-                    "href": "/things{?offset,limit,format,sort_by,sort_order}",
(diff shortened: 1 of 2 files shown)


farshidtz (changes requested):
This is a big feature drop hidden inside an assertion. It doesn't put the assertion at risk because other implementations implement the other methods prescribed in the same assertion.

IMO, this should be a separate assertion and have two separate backing implementations.

PR #493, opened 2023-05-22
Resolve exploration-server-http-alternate-language

mmccool:
See PR #492 - exploration-server-http-alternate-language did not actually get enough implementations to resolve its at-risk status.  This PR demotes it to an informative statement but adds a similar reference to the TD 1.1 spec defining the process by which the default language for TDDs can be changed as the related (demoted) assertion for TDDs.

***
<a href="https://pr-preview.s3.amazonaws.com/mmccool/wot-discovery/pull/493.html" title="Last updated on May 24, 2023, 12:33 PM UTC (4996db3)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-discovery/493/0bf3c7b...mmccool:4996db3.html" title="Last updated on May 24, 2023, 12:33 PM UTC (4996db3)">Diff</a>

PR #489, opened 2023-05-19
add reference to IoT'16 paper

mmccool:
Resolves #147

***

### :boom: Error: 500 Internal Server Error :boom: ###

[PR Preview](https://github.com/tobie/pr-preview#pr-preview) failed to build. _(Last tried on May 19, 2023, 5:30 PM UTC)_.

<details>
<summary>More</summary>

PR Preview relies on a number of web services to run. There seems to be an issue with the following one:

:rotating_light: [Spec Generator](https://www.w3.org/2015/labs/) - Spec Generator is the web service used to build specs that rely on ReSpec.

:link: [Related URL](https://labs.w3.org/spec-generator/?type=respec&url=https%3A%2F%2Fraw.githubusercontent.com%2Fmmccool%2Fwot-discovery%2Fdd7dff4cf073dc195b648cde77c2e82c25afb57a%2Findex.html%3FisPreview%3Dtrue)

```

😭  Sorry, there was an error generating the HTML. Please report this issue!
Specification: http://labs.w3.org/spec-generator/uploads/iT4Wzz/index.html?isPreview=true%3FisPreview%3Dtrue&publishDate=2023-05-19
ReSpec version: 34.1.1
File a bug: https://github.com/w3c/respec/
Error: Error: Evaluation failed: Timeout: document.respec.ready didn't resolve in 28279ms.
    at ExecutionContext._ExecutionContext_evaluate (/u/spec-generator/node_modules/puppeteer-core/lib/cjs/puppeteer/common/ExecutionContext.js:229:15)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async ExecutionContext.evaluate (/u/spec-generator/node_modules/puppeteer-core/lib/cjs/puppeteer/common/ExecutionContext.js:107:16)
    at async generateHTML (/u/spec-generator/node_modules/respec/tools/respecDocWriter.js:221:12)
    at async toHTML (/u/spec-generator/node_modules/respec/tools/respecDocWriter.js:92:18)
    at async Object.generate [as respec] (file:///u/spec-generator/generators/respec.js:15:44)
    at async file:///u/spec-generator/server.js:252:48

```

_If you don't have enough information above to solve the error by yourself (or to understand to which web service the error is related to, if any), please [file an issue](https://github.com/tobie/pr-preview/issues/new?title=Error%20not%20surfaced%20properly&body=See%20w3c/wot-discovery%23489.)._
</details>

PR #494, opened 2023-05-23
Interim 2023 Implementation Report Update 

mmccool:
No  changes to main document, just to testing results and implementation report.
Cleans up the following still at-risk assertions (WIP; a check means it is done):
- [x] `introduction-dns-sd-txt-record-udp` - this was introduced when the table was split up into TCP and UDP. Previously there was one assertion for the TXT record, but (consistent with the IANA registration) it was split into two, one each for TCP and UDP.  The TCP one kept the original assertion id, then this one was added for UDP. I'm sure implementors of the UDP discovery mechanism did do this (since otherwise it would not work...), but technically we need to ask people who worked on UDP discovery to report on this new assertion, or have some policy to decide on how to interpret the existing results for it. One solution would be to "clone" the results for the original assertion that was split up, but I think we would need a group resolution to adopt that policy.
    - Resolution: New results submitted: https://github.com/w3c/wot-testing/pull/576. Note this also removed an (incorrect) result from node-wot for DNS-SD using TCP but it still passes.
- [x] `introduction-did-service-context `- somehow this did not make it into the template, but is for the addition of an `@context` link to the DID results (since they are JSON-LD). We have to get implementors who worked on DID to update their results to include the `@context`, and report on it.
    - Resolution: New test results submitted: https://github.com/w3c/wot-testing/pull/577
- [x] `exploration-secboot-oauth2-flows` - https://github.com/w3c/wot-discovery/pull/485 was closed without merging (as it was considered to be broken) but we still have to figure out how to resolve that at-risk item without breaking other things.
    - Resolution: https://github.com/w3c/wot-discovery/pull/495 makes this informative, but for consistency also makes `exploration-secboot-auth` informative and generalizes `exploration-secboot-401`.  Instead of specifying response codes, these now rely on the relevant external specifications by way of the IANA registrations.  Note that all three assertions were at risk going into CR, and these modifications do not impact implementations.
- [x] `exploration-server-http-alternate-language` - It looks like this at-risk assertion for TD Server did NOT pass, after all.
    - Resolution: https://github.com/w3c/wot-discovery/pull/493 makes this informative.  However, we will wait to merge this until the main call as it changes the document.

***

### :boom: Error: 500 Internal Server Error :boom: ###

[PR Preview](https://github.com/tobie/pr-preview#pr-preview) failed to build. _(Last tried on May 23, 2023, 1:37 PM UTC)_.

<details>
<summary>More</summary>

PR Preview relies on a number of web services to run. There seems to be an issue with the following one:

:rotating_light: [Spec Generator](https://www.w3.org/2015/labs/) - Spec Generator is the web service used to build specs that rely on ReSpec.

:link: [Related URL](https://labs.w3.org/spec-generator/?type=respec&url=https%3A%2F%2Fraw.githubusercontent.com%2Fw3c%2Fwot-discovery%2Fe156883aebd06f859077000d881b3e8c1f93d219%2Findex.html%3FisPreview%3Dtrue)

```

😭  Sorry, there was an error generating the HTML. Please report this issue!
Specification: http://labs.w3.org/spec-generator/uploads/rnBUKd/index.html?isPreview=true%3FisPreview%3Dtrue&publishDate=2023-05-23
ReSpec version: 34.1.1
File a bug: https://github.com/w3c/respec/
Error: Error: Evaluation failed: Timeout: document.respec.ready didn't resolve in 28394ms.
    at ExecutionContext._ExecutionContext_evaluate (/u/spec-generator/node_modules/puppeteer-core/lib/cjs/puppeteer/common/ExecutionContext.js:229:15)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async ExecutionContext.evaluate (/u/spec-generator/node_modules/puppeteer-core/lib/cjs/puppeteer/common/ExecutionContext.js:107:16)
    at async generateHTML (/u/spec-generator/node_modules/respec/tools/respecDocWriter.js:221:12)
    at async toHTML (/u/spec-generator/node_modules/respec/tools/respecDocWriter.js:92:18)
    at async Object.generate [as respec] (file:///u/spec-generator/generators/respec.js:15:44)
    at async file:///u/spec-generator/server.js:252:48

```

_If you don't have enough information above to solve the error by yourself (or to understand to which web service the error is related to, if any), please [file an issue](https://github.com/tobie/pr-preview/issues/new?title=Error%20not%20surfaced%20properly&body=See%20w3c/wot-discovery%23494.)._
</details>


mmccool:
Preview is still broken, unfortunately, due to ReSpec header bug in base file.   Please look at the source diff.

mmccool:
All at-risk items have now been resolved.

static.html
```diff
@@ -3369,10 +3366,25 @@ <h1 id="title" class="title">Web of Things (WoT) Discovery</h1>
         
         <p>Special thanks to Arne Broering, Soumya Kanti Datta, and Christian Bonnet
            for their survey work on existing discovery mechannisms in IoT.
```
JKRhb, line 3368: As already mentioned in https://github.com/w3c/wot-discovery/pull/489/files#r1204112219, there is a small typo in this line: 

```suggestion
           for their survey work on existing discovery mechanisms in IoT.
```

mmccool: Is in the report, to fix this I will have to correct it in the template.  Will do so in the next PR.

PR #495, opened 2023-05-23
Resolve security bootstrapping - round 2

mmccool:
- As discussed in https://github.com/w3c/wot-discovery/pull/485, make changes to resolve at-risk status of exploration-secboot-oauth2-flows
- However, just removing this assertion does not work, as there are dependencies with two other assertions, exploration-secboot-401 and exploration-secboot-auth
- All three assertions were at risk going into CR

Resolution:
    - Reword (generalize) exploration-secboot-401 to rely on the standards for the related authentication schemes, rather than mentioning specific response codes.  This does not actually change the response codes that would be used in practice, so would not impact implementations.  By not mentioning specific response codes however we can avoid a broken dependence if exploration-secboot-oauth2-flows was just made informative.
    - Make both exploration-secboot-auth and exploration-secboot-oauth2-flows informative; they are both technically unecessary if we are relying on the standards related to the relevant security schemes.  Add a comment about this reliance so it is clear.

Note: I also removed a mention of the OAuth2 device flow since this is no longer supported in TD 1.1.

***
<a href="https://pr-preview.s3.amazonaws.com/mmccool/wot-discovery/pull/495.html" title="Last updated on May 24, 2023, 12:55 PM UTC (0b9490c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/wot-discovery/495/01e8c06...mmccool:0b9490c.html" title="Last updated on May 24, 2023, 12:55 PM UTC (0b9490c)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -1035,22 +1035,23 @@ img.wot-diagram {
                 appropriate if the TD served as a response does not contain and cannot be used to infer 
                 Personally Identifiable Information; see <a href="#privacy-considerations"></a>.
                 </li>
-                <li><span class="rfc2119-assertion" id="exploration-secboot-auth">
+                <li><!-- <span class="rfc2119-assertion" id="exploration-secboot-auth"> -->
                 If security bootstrapping is enabled on an exploration service
                 using one of the following IANA-registered
-                HTTP Authentication Schemes: Basic, Bearer, or Digest,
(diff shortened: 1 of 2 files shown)


benfrancis (commented):
@mmccool This is much better, thank you.

PR #511, opened 2023-07-27
Update discovery-ontology.ttl

AndreaCimminoArriaga:
Ontology updated with WotDirectory, WotThing types (subclasses of did:Service)

### DIFF
--- a/context/discovery-ontology.ttl
+++ b/context/discovery-ontology.ttl
@@ -118,7 +118,8 @@ discovery:WotThingDIDService rdf:type owl:Class ;
 ### discovery:WotDirectory
 discovery:WotDirectoryDIDService rdf:type owl:Class ;
 
-	rdfs:comment "Class for DID documents";
+	rdfs:comment "Class used in DID documents that labels a URL for a service returning a TD that must describe a TD Directory";
+
         rdfs:subClassOf did:Service ;
 	rdfs:label "WotDirectoryDIDService".
 
--- a/context/discovery-ontology.ttl
+++ b/context/discovery-ontology.ttl
@@ -120,7 +120,8 @@ discovery:WotDirectoryDIDService rdf:type owl:Class ;
 
 	rdfs:comment "Class for DID documents";
         rdfs:subClassOf did:Service ;
-	rdfs:label "WotDirectory".
+	rdfs:label "WotDirectoryDIDService".
+
 
 #################################################################
 # Instances
--- a/context/discovery-ontology.ttl
+++ b/context/discovery-ontology.ttl
@@ -8,6 +8,7 @@
 @prefix vann: <http://purl.org/vocab/vann/> .
 @prefix discovery: <https://github.com/w3c/wot-discovery#> .
 @prefix td: <https://www.w3.org/2019/wot/td#> .
+@prefix did: <https://www.w3.org/ns/did#> .
 @prefix schema: <http://schema.org/> .
 @base <https://github.com/w3c/wot-discovery#> .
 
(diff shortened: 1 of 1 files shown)


mmccool:
You didn't join last meeting and I realized I should have added some comments.  We need a separate ontology file for this, not changes to the current one.  I don't want the DID file to have to pull in all the other definitions, and don't want to have to change the URL or content in DID documents from what's currently in the PR against the DID documents.

mmccool:
It's possible I don't understand what you did, though... if you can't join the meetings, can you explain in comments here?

AndreaCimminoArriaga:
The ontology contains the definitions of all the concepts related to the wot discovery, as far as I know this is the best way to proceed.  I created two new types discovery:WotThing and discovery:WotDirectory both subclasses of did:Service (this is how the did document specifies it should be extended with new type of services). 

On the other hand, in order to not need to change anything in the json-ld payloads the key is to have separate context files (that's why PR #512 creates a different context file). Following this approach, the discovery (not did) payloads may keep using the context as they are doing now (no need to change anything), the did payloads can use (as they do now) the WotThing and WotDirectory types (so anything needs to be changed). The only thing to take into account is that the did payloads must use the context file from PR #512.

context/discovery-ontology.ttl
```diff
@@ -105,6 +106,18 @@ discovery:ThingDirectory rdf:type owl:Class ;
 discovery:ThingCollection rdf:type owl:Class ;
 	rdfs:label "ThingCollection".
 
+### discovery:WotThing
+discovery:WotThing rdf:type owl:Class ;
```
mmccool, line 110:
```suggestion
discovery:WotThingDIDService rdf:type owl:Class ;

```

context/discovery-ontology.ttl
```diff
@@ -105,6 +106,18 @@ discovery:ThingDirectory rdf:type owl:Class ;
...
 	rdfs:label "ThingCollection".
 
+### discovery:WotThing
+discovery:WotThing rdf:type owl:Class ;
+	rdfs:comment "Class for DID documents";
+	rdfs:subClassOf did:Service ;
+	rdfs:label "WotThing".
+
+### discovery:WotDirectory
+discovery:WotDirectory rdf:type owl:Class ;
```
mmccool, line 116:
```suggestion
discovery:WotDirectoryDIDService rdf:type owl:Class ;

```

context/discovery-ontology.ttl
```diff
@@ -105,6 +106,18 @@ discovery:ThingDirectory rdf:type owl:Class ;
...
+	rdfs:comment "Class for DID documents";
```
mmccool, line 117:
```suggestion
	rdfs:comment "Class for DID documents that labels URL for service returning a TD that must be a TD Directory";

```

context/discovery-ontology.ttl
```diff
@@ -105,6 +106,18 @@ discovery:ThingDirectory rdf:type owl:Class ;
 discovery:ThingCollection rdf:type owl:Class ;
 	rdfs:label "ThingCollection".
 
+### discovery:WotThing
+discovery:WotThing rdf:type owl:Class ;
+	rdfs:comment "Class for DID documents";
```
mmccool, line 111:
```suggestion
	rdfs:comment "Class for DID documents that labels URL for a service returning a TD";

```

context/discovery-ontology.ttl
```diff
@@ -105,6 +106,18 @@ discovery:ThingDirectory rdf:type owl:Class ;
...
+discovery:WotThing rdf:type owl:Class ;
+	rdfs:comment "Class for DID documents";
+	rdfs:subClassOf did:Service ;
+	rdfs:label "WotThing".
+
+### discovery:WotDirectory
+discovery:WotDirectory rdf:type owl:Class ;
+	rdfs:comment "Class for DID documents";
+        rdfs:subClassOf did:Service ;
+	rdfs:label "WotDirectory".
```
mmccool, line 119:
```suggestion
	rdfs:label "WotDirectoryDIDService".

```

PR #513, opened 2023-08-31
Add wot-discovery to the list of the implementations

lu-zero:


egekorkan (approved):

implementations/README.md
```diff
@@ -20,10 +20,13 @@ Self-description also needs to be tested:
 
 In addition, the following introduction mechanisms have been implemented and tested:
 * Direct URL
-* [Hitachi - DNS-SD/mDNS](implementations/hitachi_intro.md)
+* DNS-SD/mDNS
+    * [Hitachi - DNS-SD/mDNS](implementations/hitachi_intro.md)
```
danielpeintner, line 24: 2 things, one related to this PR and the other one not:
* maybe change [Hitachi - DNS-SD/mDNS] to [Hitachi] only since it is duplicated information
* the linked MD file is not reachable

k-toumura: could we use `https://github.com/w3c/wot-testing/blob/main/events/2021.03.Online/TDs/Hitachi/README.md` as link for the information?
```suggestion
    * [Hitachi](https://github.com/w3c/wot-testing/blob/main/events/2021.03.Online/TDs/Hitachi/README.md)
```

but the implementation is only available on plugfest and not publicly accessible.  So, we can just remove this entry.

PR #514, opened 2023-09-03
Initial survey of other discovery mechanisms

k-toumura:
This is the first survey (draft) of discovery mechanisms in other platform/protocol/ecosystems..


mmccool:
In Discovery call Nov 6, agreed to mark as draft, but merge and reorganize.

PR #522, opened 2023-10-16
Update Resources

mmccool:
- Move things into directories by category
- Add Thing Model for TDD
- Add JSON Schema to TD Extensions
- Try to make naming conventions consistent with TD spec


egekorkan (approved):
Just that leftmost part of the publication mapping table should be updated once the resources are published at wot-resources?

mmccool:
leftmost part will be updated to actual URLs later - for now interpreted as relative URLs to the location of the README.

PR #523, opened 2023-11-08
Update static version for REC publication

mmccool:
Update static version for REC publication

- Re-export from ReSpec
- Re-run htmltidy

Purpose: update link to W3C Process document (should be Nov 3).  

However, the above did not change anything except the publication date...  and also that is wrong, ReSpec is providing the date as October 24, as opposed to today. (Update: fixed).


egekorkan:
As seen in https://github.com/w3c/wot-discovery/blob/rec-prep/publication/6-rec/Overview.html#L1321-L1322 , the respec is not updating the process. In the TD call, we have learned that respec is not updated yet but should be tomorrow. So we should not merge now.

Issue #533, opened 2023-12-11
Update HTML for Discovery ontology


mmccool:
- a temporary HTML file representing the discovery ontology has been created: https://github.com/w3c/wot-resources/blob/main/discovery/v1/ontology/discovery-ontology.html
- the source is here: https://github.com/w3c/wot-discovery/blob/main/publication/6-resources/ontology/discovery-ontology.html
- available at official URL: https://www.w3.org/2022/wot/discovery-ontology

To do: 
- [ ] Set up the scripts to generate the HTML for the discovery ontology (copy and modify scripts from TD repo) - PR into wot-discovery repo
- [ ] Update the HTML file to reflect the actual content of the discovery ontology - PR into wot-discovery repo
- [ ] Replace the HTML resources in wot-resources (the publication is by redirection, so this will update the published resource at the official URL) - PR into wot-resources repo

PR #535, opened 2024-01-08
Add rendering script for discovery ontology

JKRhb:
This PR adds a mechanism for rendering the discovery ontology as an HTML document, similar to the one in the TD repository. When merged, this PR will tick the first two boxes of https://github.com/w3c/wot-discovery/issues/533.

While the result should already contain all terms from the `.ttl` file, some improvements are still needed. For instance, all terms are currently missing a description. Furthermore, the rendered document needs more surrounding text, e.g., an introduction and maybe an expanded abstract.

Note that I had to make some minor adjustments to the `.ttl` file (adding a `preferredNamespacePrefix` and a `preferredNamespaceUri`). I also refactored and simplified the base URI usage a bit to align it with the `.ttl` files in the TD repository.


egekorkan:
@JKRhb could you also add a readme just to document how it differs from the TD spec pipeline (other than a different template)

context/discovery-ontology.template.html
```diff
@@ -0,0 +1,49 @@
...
+        <p>
+            This document provides an ontology describing discovery mechanisms
+            for the Web of Things (WoT).
+            These mechanisms include Thing Description Directories (TDDs) as
+            well as Thing Links.
+        </p>
+    </section>
+    <section id='sotd'>
+        <p>
+            TODO
```
mmccool, line 37:
```suggestion
            Version 1.0.1.
            See [[#changelog]].
```

context/discovery-ontology.template.html
```diff
@@ -0,0 +1,49 @@
...
+        </p>
+    </section>
+    <section>
+        <h2>Introduction</h2>
+    </section>
+
+    <!-- axioms rendered from RDF definitions -->
+    %s
+
```
mmccool, line 46: <section>
        <h2>Change Log</h2>
            <p>
            The following lists the changes to the ontology for each version.  Semantic versioning is used, so major version changes are breaking,
minor version changes are substantial but non-breaking, and bug fix versions are small and non-breaking.
            </p>
            <dl>
            <dt>Version 1.0.1</dt><dd>Script used to render HTML from TTL.  TTL updated to include annotations and comments, and defining default namespace.
            </dd>
            <dt>Version 1.0.0</dt><dd>Initial release.  Placeholder HTML only.
            </dd>
            </dl>

JKRhb:
```suggestion
    <section id="changelog">
        <h2>Change Log</h2>
            <p>
            The following lists the changes to the ontology for each version.  Semantic versioning is used, so major version changes are breaking,
minor version changes are substantial but non-breaking, and bug fix versions are small and non-breaking.
            </p>
            <dl>
            <dt>Version 1.0.1</dt><dd>Script used to render HTML from TTL.  TTL updated to include annotations and comments, and defining default namespace.
            </dd>
            <dt>Version 1.0.0</dt><dd>Initial release.  Placeholder HTML only.
            </dd>
            </dl>
```

PR #536, opened 2024-01-16
Render changelog in discovery ontology from RDF

JKRhb:
As discussed in the last discovery call, this PR tries to add a changelog to the `.ttl` file and render it dynamically in the HTML document from the `template.sparql` file. I am not sure yet if this is the best or even a good approach, so feedback is greatly appreciated :)

### DIFF
--- a/context/discovery-ontology.template.html
+++ b/context/discovery-ontology.template.html
@@ -44,17 +44,19 @@
 
     <!-- axioms rendered from RDF definitions -->
     %s
+
     <section id="changelog">
         <h2>Change Log</h2>
             <p>
-            The following lists the changes to the ontology for each version.  Semantic versioning is used, so major version changes are breaking,
-minor version changes are substantial but non-breaking, and bug fix versions are small and non-breaking.
+                The following lists the changes to the ontology for each
+                version.
+                Semantic versioning is used, so major version changes are
+                breaking, minor version changes are substantial but
+                non-breaking, and bug fix versions are small and non-breaking.
             </p>
             <dl>
-            <dt>Version 1.0.1</dt><dd>Script used to render HTML from TTL.  TTL updated to include annotations and comments, and defining default namespace.
-            </dd>
-            <dt>Version 1.0.0</dt><dd>Initial release.  Placeholder HTML only.
-            </dd>
+                <!-- Changelog rendered from .ttl file -->
+                %s
             </dl>
      </section>
 </body>
--- a/context/discovery-ontology.template.html
+++ b/context/discovery-ontology.template.html
@@ -7,8 +7,12 @@
     <script src='https://www.w3.org/Tools/respec/respec-w3c' class='remove'></script>
     <script class='remove'>
         var respecConfig = {
-            specStatus: "REC",
+            specStatus : "ED",
+            copyrightStart : 2017,
             latestVersion: null,
+            group: "wg/wot",
+            wgPublicList : "public-wot-wg",
+            edDraftURI: "https://w3c.github.io/wot-discovery/context/discovery-ontology",
             editors: [{
                 name: "Andrea Cimmino"
             }, {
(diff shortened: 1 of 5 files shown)


JKRhb:
Thanks to the information on versioning provided by @mahdanoura in the [TD repo](https://github.com/w3c/wot-thing-description/blob/2f360cc9309b472d33f7c5ad8f463a13466a3419/VERSIONING.md#ontologies), I now tried to adjust the rendering script in a way that enables us to generate a changelog automatically.

For that purpose, I now tried to use `owl:versionInfo` as well as `vann:changes` to extract information relevant for the changelog and add it to the HTML version of the ontology.

Using `owl:priorVersion`, it should then be possible to retrieve the respective previous version of the ontology, render the changelog information there, and then continue recursively with the previous ontology version until no `owl:priorVersion` is defined. This logic is currently not yet implemented because I wanted to discuss the general direction first, but if you think this is reasonable then I would try to come up with a completed version.

context/template.sparql
```diff
@@ -48,6 +51,17 @@ template :axioms(?prefix) {
...
 
+template :changelog(?prefix) {
+    format { "<dt>%s</dt>" ?versionInfo }
+    format { "<dd>%s</dd>" ?changes }
+} where {
+    ?o a owl:Ontology ;
+         owl:versionInfo ?versionInfo ;
+         owl:priorVersion ?priorVersion ;
+         vann:changes ?changes .
+    # TODO: Call template with priorVersion URL
```
JKRhb, line 62: After trying out a number of different strategies, I haven't really found a working approach to recursively call this template while using `priorVersion` as the new reference point. Do you maybe have an idea, @mahdanoura, how this could be done and whether this is possible at all using Sparkle Templates?

JKRhb: I think I've now got this to work :) This required some additional adjustments in the JS-based render script, but eventually I managed to render the changelog from `owl:priorVersion`, `owl:versionInfo`, and `vann:changes`.

However, as there is no version 1.0.0 published yet, it is currently not possible to refer to the previous version. Thus, we might want to publish the placeholder HTML first or make this version the actual version 1.0.0.

mahdanoura: > After already trying out a couple of different strategies in the meantime, I haven't really found a working approach to recursively call this template while using `priorVersion` as the new reference point. Do you maybe have an idea, @mahdanoura, how this could be done and whether this is possible at all using SPARQL Templates?

To the best of my knowledge, it is not possible to achieve such a recursive behaviour with SPARQL templates.

context/discovery-ontology.template.html
mahdanoura, line 1: As an idea, maybe we keep the metadata of the ontology HTML pages similar to the TD specification adding a `This Version`, `Latest published version`, `History`, and `Commits` (if we can use the tags to filter on specific commits) as links to previous ontologies for easier access.

mahdanoura (approved):

context/discovery-ontology.template.html
JKRhb: Thank you, @mahdanoura, I tried to incorporate this in 7cab1d518b07420a2c9ec8b336f158c103740a13, which should also add a correct link to the "Editor's Draft" of the ontology alongside some additional metadata. Regarding the `History`/`Commits`, it seems as if it is not possible to filter for a specific tag or commit range, though, according to the [ReSpec documentation](https://respec.org/docs/#github), which is a bit unfortunate :/

mmccool:
HTML Rendering: https://htmlpreview.github.io/

context/discovery-ontology.ttl
```diff
@@ -14,7 +14,10 @@
 : rdf:type owl:Ontology ;
 	dc:creator "Andrea Cimmino" ;
 	dc:creator "Farshid Tavakolizadeh" ;
-	owl:versionInfo "0.1.2" ;
+	owl:versionInfo "1.0.1" ;
```
mmccool, line 17:
```suggestion
	owl:versionInfo "1.0.0-a2" ;

```

JKRhb: I've applied the new version now and also adjusted the `priorVersion` URL so that it actually fetches the `.ttl` file from GitHub.

PR #537, opened 2024-01-16
Expand `rdfs:comment`s in Discovery ontology

JKRhb:
This PR will eventually tick the third box of #533.

Currently, it is still WIP.


mmccool:
HTML rendering: https://htmlpreview.github.io/?https://github.com/w3c/wot-discovery/blob/b3b3df907ad5b7a71fa08f7e24c36f80bea571ed/context/discovery-ontology.html

